PR #20, opened 2014-04-23
Implement stripping

mmun:
The parser now respects the handlebars whitespace control flags. Whitespace is removed from the AST at parse time. Any unnecessary text nodes are removed. This works nicely with single element programs. Instead of writing

``` handlebars
{{#each}}<li>foo</li>{{/each}}
```

you can write

``` handlebars
{{#each~}}
  <li>foo</li>
{{~/each}}
```

and still generate an optimized fragment

``` javascript
function build(dom) {
  var el0 = dom.createElement('li');
  dom.appendText(el0, 'foo');
  return el0;
}
```

### DIFF
--- a/lib/htmlbars/ast.js
+++ b/lib/htmlbars/ast.js
@@ -1,12 +1,24 @@
 import AST from "handlebars/compiler/ast";
 
-export var ProgramNode = AST.ProgramNode;
-export var BlockNode = AST.BlockNode;
 export var MustacheNode = AST.MustacheNode;
 export var SexprNode = AST.SexprNode;
 export var HashNode = AST.HashNode;
 export var IdNode = AST.IdNode;
 export var StringNode = AST.StringNode;
+
+export function ProgramNode(statements, strip) {
+  this.type = 'program';
+  this.statements = statements;
+  this.strip = strip;
+}
+
+export function BlockNode(mustache, program, inverse, strip) {
+  this.type = 'block';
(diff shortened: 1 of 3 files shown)


oligriffiths:
Hey, I see this works for `{{#each~}}`, however it doesn't seem to work for an `if` statement:

works as expected:
```
{{#each titleParts as |part|~}}
    {{part.value}}
{{~/each}}
```

spaces appear:
```
{{#each titleParts as |part|~}}
    {{#if part.highlight}}
        <span class="highlight">
            {{part.value}}
        </span>
    {{else}}
        {{part.value}}
    {{/if}}
{{~/each}}
```

Any advice?

Thanks

Cheralyn:
This is an automatically generated message.

coferc@yahoo-inc.com is no longer with Yahoo! Inc.

Your message will not be forwarded.

If you have a sales inquiry, please email yahoosales@yahoo-inc.com and someone will follow up with you shortly.

If you require assistance with a legal matter, please send a message to legal-notices@yahoo-inc.com

Thank you!

oligriffiths:
It appears that 
```
{{#each titleParts as |part|~}}
    {{~#if part.highlight~}}
        <span class="highlight">{{part.value}}</span>
    {{~else~}}
        <span class="no-highlight">{{part.value}}</span>
    {{/if}}
{{~/each}}
``` 
does the trick

PR #129, opened 2014-11-14
[RFC] Use latest module transpiler

eventualbuddha:
This is a work in progress. Some things in the build system, such as CommonJS output, we did not replicate in this branch. What are the intended build artifacts for htmlbars? What is intended to be run both in the browser and node?

cc @chadhietala


mmun:
Anyone interested in migrating to Esperanto?

Issue #87, opened 2014-09-04
createElement should accept a hash of attributes


mmun:
This is required for compatibility with IE6 to abstract away these kinds of hacks:

https://github.com/emberjs/ember.js/blob/0e7a08ea93f7e7b4fbecdb474854584ecc0b2cfe/packages/ember-views/lib/system/render_buffer.js#L454-L461

philn5d:
Can have a variant on create element that just takes the tag and shoves into element. It looks like createElement does not need to change, but just set up the tag to pass in before calling the existing.

    function createTagRaw(tag, attr){
        //do loop to make IE6 tag
        return tagForAncientBrowserThatIsNoLongerSupportedOrRecommended;
    }

But honestly, IMO why slow down or clutter createElement for the sake of supporting a browser that's 5 versions out of date? Just saying...

philn5d:
Better yet, how about an IE6 shim? Could create as an add-in to support IE6. Would definitely be more efficient than one-offs for each function to support IE6. In other words, if you care about supporting it, require('ie6shim') and have that override each method that needs to be altered to support it. Probably a better pattern anyways instead of making each method inefficient and cluttered.

mixonic:
You cannot shim support for setting a name after input creation,
unfortunately. By the time `createElement('input')` has been called it
is already too late to set a name. You need to know the name at
`createElement` time.

On Sat, Feb 14, 2015, at 05:25 PM, philn5d wrote:
> Better yet, how about an IE6 shim? Could create as an add-in to
> support IE6. Would definitely be more efficient than one-offs for each
> function to support IE6. In other words, if you care about supporting
> it, require('ie6shim') and have that override each method that needs
> to be altered to support it. Probably a better pattern anyways instead
> of making each method inefficient and cluttered.


> — Reply to this email directly or view it on GitHub[1].

>



Links:

  1. https://github.com/tildeio/htmlbars/issues/87#issuecomment-74394614

PR #213, opened 2014-12-15
[FEATURE] Allow createElement to take a hash of attributes

chadhietala:
Should close out #87.


mmun:
Should we keep this open?

mixonic:
I'll close for now, we can re-open if need be. I still think an all-at-once helper makes sense, but not one that bothers with the build-attrs-in-the-string method. It would just be sugar on top of `createElement` and `setAttribute`, and help build a dsl for writing DOM by hand.

Issue #229, opened 2014-12-27
IE test failures


oneeman:
As pointed out by @johanneswuerbach in https://github.com/tildeio/htmlbars/issues/220, the current build has failing tests with IE.  I'm working on getting fixing that so they at least all pass with IE 11, which I have on my windows VM.

PR #230, opened 2014-12-27
Fix test failures on Internet Explorer (#229)

oneeman:
Fixes https://github.com/tildeio/htmlbars/issues/229.

-- Convert value for setAttribute() to string
-- Make repairClonedNode() safe to call on uncloned nodes (e.g. first rendering)

Tested with IE 11, and I expect based on what was failing before that it will also take care of IE 9 and 10, but let's see what the saucelabs tests say.  My first approach was to not call `repairClonedNode()` the first time (by modifying the code for `template.render()`), but it made a different test fail, and having the dom helper handle it seems more appropriate anyway.

The added tests were not strictly speaking required, since there were other failing tests, but these make the failure's cause explicit.

[EDIT: passes IE 8+]


mixonic:
Mucho gracias.

Issue #224, opened 2014-12-23
Namespaced attributes


salzhrani:
I am willing to take a stab at it if no major changes are required.

PR #235, opened 2015-01-01
Implement #224 

salzhrani:
:)

### DIFF
--- a/packages/htmlbars-compiler/lib/utils.js
+++ b/packages/htmlbars-compiler/lib/utils.js
@@ -9,3 +9,19 @@ export function processOpcodes(compiler, opcodes) {
     }
   }
 }
+
+// ref http://dev.w3.org/html5/spec-LC/namespaces.html
+var defaultNamespaces = {
+  html: 'http://www.w3.org/1999/xhtml',
+  mathml: 'http://www.w3.org/1998/Math/MathML',
+  svg: 'http://www.w3.org/2000/svg',
+  xlink: 'http://www.w3.org/1999/xlink',
+  xml: 'http://www.w3.org/XML/1998/namespace'
+};
+
+export function getNamespace(attrName) {
+  var parts = attrName.split(':');
+  if (parts.length > 1) {
+    return defaultNamespaces[parts[0]];
+  }
+}
--- a/packages/morph/lib/attr-morph.js
+++ b/packages/morph/lib/attr-morph.js
@@ -14,27 +14,41 @@ function updateAttribute(value) {
   }
 }
 
-function AttrMorph(element, attrName, domHelper) {
+function updateAttributeNS(value) {
+  if (value === null) {
+    this.domHelper.removeAttribute(this.element, this.attrName);
+  } else {
+    this.domHelper.setAttributeNS(this.element, this.namespace, this.attrName, value);
+  }
+}
+
+function AttrMorph(element, attrName, domHelper, namespace) {
   this.element = element;
   this.domHelper = domHelper;
+  this.namespace = namespace || null;
   this.escaped = true;
 
   var normalizedAttrName = normalizeProperty(this.element, attrName);
-  if (element.namespaceURI === svgNamespace || attrName === 'style' || !normalizedAttrName) {
+  if (this.namespace) {
+    this._update = updateAttributeNS;
     this.attrName = attrName;
-    this._update = updateAttribute;
   } else {
-    this.attrName = normalizedAttrName;
-    this._update = updateProperty;
+    if (element.namespaceURI === svgNamespace || attrName === 'style' || !normalizedAttrName) {
+      this.attrName = attrName;
+      this._update = updateAttribute;
+    } else {
+      this.attrName = normalizedAttrName;
+      this._update = updateProperty;
+    }
   }
 }
 
 AttrMorph.prototype.setContent = function (value) {
   if (this.escaped) {
     var sanitized = sanitizeAttributeValue(this.element, this.attrName, value);
-    this._update(sanitized);
+    this._update(sanitized, this.namespace);
   } else {
-    this._update(value);
+    this._update(value, this.namespace);
   }
 };
 
--- a/packages/morph/tests/dom-helper-test.js
+++ b/packages/morph/tests/dom-helper-test.js
@@ -47,6 +47,21 @@ test('#setAttribute', function(){
   equalHTML(node, '<input disabled="false">');
 });
 
+test('#setAttributeNS', function(){
+  var node = dom.createElement('svg');
+  dom.setAttributeNS(node, 'http://www.w3.org/1999/xlink', 'xlink:href', 'super-fun');
+  // chrome adds (xmlns:xlink="http://www.w3.org/1999/xlink") property while others don't
+  // thus equalHTML is not useful
+  var el = document.createElement('div');
+  el.appendChild(node);
+  // phantomjs omits the prefix, thus we can't find xlink:
+  ok(el.innerHTML.indexOf('href="super-fun"') > 0);
+  dom.setAttributeNS(node, 'http://www.w3.org/1999/xlink', 'href', null);
+  
+  ok(el.innerHTML.indexOf('href="null"') > 0);
+
+});
+
 test('#getElementById', function() {
   var parentNode = dom.createElement('div'),
       childNode = dom.createElement('div');
@@ -121,6 +136,14 @@ test('#setProperty', function(){
   node = dom.createElement('svg');
   dom.setProperty(node, 'viewBox', '0 0 0 0');
   equalHTML(node, '<svg viewBox="0 0 0 0"></svg>');
+  
+  dom.setProperty(node, 'xlink:title', 'super-blast','http://www.w3.org/1999/xlink');
+  // chrome adds (xmlns:xlink="http://www.w3.org/1999/xlink") property while others don't
+  // thus equalHTML is not useful
+  var el = document.createElement('div');
+  el.appendChild(node);
+  // phantom js omits the prefix so we can't look for xlink:
+  ok(el.innerHTML.indexOf('title="super-blast"') > 0 );
 });
 
 test('#addClasses', function(){
--- a/packages/htmlbars-compiler/lib/fragment-javascript-compiler.js
+++ b/packages/htmlbars-compiler/lib/fragment-javascript-compiler.js
@@ -59,9 +59,9 @@ FragmentJavaScriptCompiler.prototype.returnNode = function() {
   this.source.push(this.indent+'  return '+el+';\n');
 };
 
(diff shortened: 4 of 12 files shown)


mmun:
Can we omit the namespace argument in the compiled JavaScript if its not used?

e.g. instead of `dom.setProperty(..., null)` and `dom.createAttrMorph(..., null)` just drop the last argument.

mixonic:
Si, I agree with @mmun. This is good groundwork for supporting the feature in Ember proper.

oneeman:
Nice.  A few thoughts:
(a) Maybe factor out the namespace detection into `html-compiler/utils.js` to make this DRYer
(b) When calling `setAttributeNS`, is the name supposed to be the full name or the local name?  The reason I ask is that when I load, e.g. `<svg xml:space="preserve"></svg>` with a browser, the attribute's `localName` is `space` whereas `name` is `xml:space`.  On the other hand, the code in this PR leads to both being `space`.  I don't know if this makes a difference as I've never used this attributes for real, but the current behavior differs from what the browsers does, so I think it's worth looking into.
(c) For the tests, it doesn't look like you're actually checking that the namespace is set correctly, since you're just looking at the resulting html.  You can instead get the attribute object and look at its properties (e.g. `namespaceURI`, `name`, `localName`, `value`)

salzhrani:
a. Done
b. this [jsbin](http://emberjs.jsbin.com/viwegaceha) shows the issue, if you do `getAttributeNS` on the `use` element it would return `null` but if you use it after applying the fix it returns the proper value.
c. I have added more assertions to check for `namespaceURI`, `name`, `localName` and `value`.

FWIW I think the feature working properly

packages/htmlbars-compiler/lib/utils.js
```diff
@@ -9,3 +9,15 @@ export function processOpcodes(compiler, opcodes) {
     }
   }
 }
+
+export function getNamespace(name) {
+	var rval = {ns: null, name: null};
```
mixonic: Please use standard two-space indentation, and avoid abbreviations (`ns`). And a map:

```
// Map values should be based on http://dev.w3.org/html5/spec-LC/namespaces.html
var defaultNamespaces = {
  xlink: {
    namespaceURI: 'http://www.w3.org/1999/xlink',
    name: 'xlink'
  }
};

export function getNamespace(attrName) {
  var parts = attrName.split(':');
  if (parts.length > 1) {
    return defaultNamespaces[parts[0]];
  }
}
```

packages/morph/lib/attr-morph.js
```diff
@@ -6,17 +6,20 @@ function updateProperty(value) {
...
-  } else {
+  } else if (namespace === null) {
     this.domHelper.setAttribute(this.element, this.attrName, value);
+  } else {
+    this.domHelper.setAttributeNS (this.element, this.namespace, this.attrName, value);
   }
 }
 
-function AttrMorph(element, attrName, domHelper) {
+function AttrMorph(element, attrName, namespace, domHelper) {
```
mixonic: Again please make `namespace` an optional last value.

packages/morph/lib/attr-morph.js
```diff
@@ -6,17 +6,20 @@ function updateProperty(value) {
   this.domHelper.setPropertyStrict(this.element, this.attrName, value);
 }
 
-function updateAttribute(value) {
+function updateAttribute(value, namespace) {
```
mixonic: Please create a separate function for updating a value with a namespace, and assign it to `this._update`. This avoids the branching in logic at update time and ensures that all `_update` functions share an argument signature.

```js
function updateAttributeNS(value) {
  if (value === null) {
    this.domHelper.removeAttribute(this.element, this.attrName);
  } else {
    this.domHelper.setAttributeNS(this.element, this.namespace, this.attrName, value);
  }
}

function AttrMorph(element, attrName, domHelper, namespace) {
  // ....
  this.namespace = namespace || null;
  // ...

  if (element.namespaceURI === svgNamespace || attrName === 'style' || !normalizedAttrName) {
    this.attrName = attrName;
    this._update = updateAttribute;
  } else {
    this.attrName = normalizedAttrName;
    if (this.namespace === null) {
      this._update = updateProperty;
    } else {
      this._update = updateAttributeNS;
    }
  }
}
```

oneeman:
@salzhrani Thanks for the demo!  It does work as is, but I was still wondering which way it should be.  So I looked at the specs, and both W3 and WhatWG say that the `name` parameter to `setAttributeNS` should be the qualified name, e.g. `xlink:href`.  The relevant specs are at:
http://www.w3.org/TR/DOM-Level-2-Core/core.html#ID-ElSetAttrNS
and 
https://dom.spec.whatwg.org/#dom-element-setattributens
https://dom.spec.whatwg.org/#validate-and-extract

So I would say to pass the qualified name instead of the local name.  I believe this would simplify the code as well, since `getNamespace` could return just the namespace uri string instead of an object with the uri and the local name.

oneeman:
Also, there's the failing tests under safari.  Maybe they're related to the `name` parameter and would get fixed when it's changed, but otherwise I think someone with a mac might need to look at this.

packages/morph/lib/attr-morph.js
oneeman: @mixonic This would lead to `updateAttribute` being used for namespaced attributes on svg-namespaced elements.  Instead, should add `namespace` as a condition on the outer `if`, move the `this.namespace` check to the outer `if`'s then clause, and leave the outer `if`'s else clause with `updateProperty`.  

I would also replace `namespace` with `attrNamespace`, just to be extra explicit, but maybe that's going too far.

salzhrani:
I have incorporated your feedback and code. @oneeman you are correct about using the qualified name with Safari, test pass now.

oneeman:
Thanks, looks good.  Sorry for going on like this, but I have two residual questions (for whomever may know the answers):
1. I noticed that the use of `createElementNS` in the dom helper is conditional on feature detection with `if (doc && doc.createElementNS)`.  Is something like this also required for `setAttributeNS`, like checking whether it's available and falling back on `setAttribute` if it's not?  
2. I wonder whether the 'xmlns' prefix should be be handled or not.  My understanding is that it sets the specified uri as the namespace for attributes with the specific prefix for all nested elements ('scoped namespace').  But the code as is doesn't do that (try `<svg xmlns:xml="http://www.w3.org/2001/XMLSchema"><text xml:space="preserve"></text></svg>`, and the namespace of the `xml:space` attribute will be `'http://www.w3.org/XML/1998/namespace'` anyway.  So I think, since its functionality is not supported, it might be better to leave it out entirely for now, but I'd like to hear what others think.  And my impression is that adding support for it later would be doable, using a similar approach to that currently used for elements' namespaces in the fragment compiler.

salzhrani:
@oneeman your input is greatly appreciated.
1. I agree and will implement if there are no other comments.
2. I wasn't sure to include all the namespaces listed, I think the major two are `xlink` and `xml`

mixonic:
The first five namepaces in http://dev.w3.org/html5/spec-LC/namespaces.html seem appropriate to support. We definitely do not want to support dynamic namespaces at this point, leaving `xmlns` off sounds like a great path.

packages/htmlbars-compiler/lib/utils.js
```diff
@@ -9,3 +9,17 @@ export function processOpcodes(compiler, opcodes) {
...
+  xlink: 'http://www.w3.org/1999/xlink',
+  xmlns: 'http://www.w3.org/2000/xmlns/',
+  xml: 'http://www.w3.org/XML/1998/namespace'
+};
+
+export function getNamespace(attrName) {
+  var parts = attrName.split(':');
+    if (parts.length > 1) {
+      return defaultNamespaces[parts[0]];
+    }
```
mixonic: Please use two space indentation.

packages/morph/lib/attr-morph.js
```diff
@@ -14,27 +14,41 @@ function updateAttribute(value) {
   }
 }
 
-function AttrMorph(element, attrName, domHelper) {
+function updateAttributeNS(value) {
+  if (value === null) {
+    this.domHelper.removeAttribute(this.element, this.attrName);
+  } else {
+    this.domHelper.setAttributeNS (this.element, this.namespace, this.attrName, value);
```
mixonic: This is meant to be `setAttributeNS(` with no space I presume.

packages/morph/lib/attr-morph.js
mixonic: > This would lead to updateAttribute being used for namespaced attributes on svg-namespaced elements.

The `namespace` of attributes should have nothing to do with the element. Why would adding another `_update` interface to `AttrMorph` change anything for svg namespaced elements?

mixonic:
@oneeman 

> 1. I noticed that the use of createElementNS in the dom helper is conditional on feature detection with if (doc && doc.createElementNS). Is something like this also required for setAttributeNS, like checking whether it's available and falling back on setAttribute if it's not?

Most of the supports-namespace tests are for building optimal code paths with less branching for browsers that don't support those features. IE8 will behave badly if you try to use SVG, and that is ok. I don't think we need that style of code here, open to ideas though.

packages/morph/lib/attr-morph.js
oneeman: Suppose you have `<svg><use xml:space={{space}}></use></svg>` or similar.  The `use` element will have the `namespaceURI` for svg, which is correct.  But it means that `element.namespaceURI === svgNamespace` in the outer `if` (in your code snippet) would evaluate as true for the attribute's morph, leading it to be set with `setAttribute` rather than `setAttributeNS`.  

At any rate, the PR's newer commits handle it correctly, so it's moot.

oneeman:
LGTM then.

packages/morph/lib/attr-morph.js
mixonic: Gotcha! Excellent, thanks for the clarification :-)

packages/morph/tests/dom-helper-test.js
```diff
@@ -121,6 +136,14 @@ test('#setProperty', function(){
   node = dom.createElement('svg');
   dom.setProperty(node, 'viewBox', '0 0 0 0');
   equalHTML(node, '<svg viewBox="0 0 0 0"></svg>');
+  dom.setProperty(node, 'xlink:title', 'super-blast','http://www.w3.org/1999/xlink');
```
mixonic: The blank line should probably go before this one and not after?

mixonic:
@salzhrani if you can address that test whitespace comment and squash these commits I will happily merge. Thanks for your hard work during the review process!

salzhrani:
@mixonic fixed and squashed. I appreciate your kindness and patience while reviewing. @oneeman you are the best :)

mixonic:
Thanks @salzhrani!

PR #236, opened 2015-01-02
Improve HTML validation and error messages thrown

oneeman:
Takes care of https://github.com/emberjs/ember.js/issues/10117 and another case where the error given wasn't ideal (it was treating the `/` as part of the attribute value and saying `Unquoted attribute value must be a single string or mustache`)

### DIFF
--- a/packages/htmlbars-compiler/tests/html-compiler-test.js
+++ b/packages/htmlbars-compiler/tests/html-compiler-test.js
@@ -152,7 +152,7 @@ test("Unquoted attribute value with multiple nodes throws an exception", functio
   QUnit.throws(function() { compile('<div \nclass\n=\n{{foo}}&amp;bar ></div>'); }, expectedError(4));
 
   function expectedError(line) {
-    return new Error("Unquoted attribute value must be a single string or mustache (line " + line + ")");
+    return new Error("Unquoted attribute value must be a single string or mustache (on line " + line + ")");
   }
 });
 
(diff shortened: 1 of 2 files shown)


rwjblue:
Awesome, thanks for tackling this @oneeman!

I can release (and update Ember) once merged.

mmun:
> A space is required between an unquoted attribute value and `/`

I don't agree with this. According to the spec, `` is equivalent to ``. The error for `` should be the same type of error as ``

mmun:
The end tag check is good. I wonder if some of these errors should be pushed up to the simple html tokenizer. They're pretty useful after all.

oneeman:
I based it on http://www.w3.org/TR/html-markup/syntax.html :
"If the value of an attribute using the unquoted attribute syntax is followed by a "/" character, then there must be at least one space character after the value and before the "/" character."

As far as handling errors on the tokenizer, I've been thinking about this too, because I expect there'll be a lot more we'll encounter in the future and at least the ones not involving mustaches should be handled there.  But since (as far as I could see) none of them were being done there for now I didn't want to start the division.  Definitely something I'd be in favor of in the future.

mmun:
@oneeman Interesting.

The tokenizer spec does not distinguish the `/` character in the unquoted attribute vlaue state: https://html.spec.whatwg.org/multipage/syntax.html#attribute-value-(unquoted)-state. Furthermore, Chrome firefox and safari will add the `/` as part of the attribute.

mmun:
The WHATWG spec also has a similar blurb:

> If an attribute using the unquoted attribute syntax is to be followed by another attribute or by the optional U+002F SOLIDUS character (/) allowed in step 6 of the start tag syntax above, then there must be a space character separating the two.

PR #237, opened 2015-01-06
Use correct namespace in parseHTML() (ember #9948)

oneeman:
This was the cause of https://github.com/emberjs/ember.js/issues/9948, and should take care of it.  The immediate problem was that the dom helper was left with the last namespace used in `build()` function.  But, more broadly, I think the appropriate namespace  for `parseHTML()` is that for the morph's contextual element (i.e. either its parent or the template's contextual element), rather than always the template's.  The two tests I added are really testing the same thing, but they cover what seem like important cases, so I figured I should include both.  

[Deleted: some mistaken tangential speculation... later findings instead:
With chrome, `parseHTML` works fine for svg stuff even if it goes down the `else` branch.  As a result, the second test I added was passing in chrome even before the fix, but it was failing in other browsers and passes after the fix.
]

### DIFF
--- a/packages/morph/lib/dom-helper.js
+++ b/packages/morph/lib/dom-helper.js
@@ -29,10 +29,6 @@ var ignoresCheckedAttribute = doc && (function(document){
   return !clonedElement.checked;
 })(doc);
 
-function isSVG(ns){
-  return ns === svgNamespace;
-}
-
 // This is not the namespace of the element, but of
 // the elements inside that elements.
 function interiorNamespace(element){
@@ -319,12 +315,7 @@ prototype.appendMorph = function(element, contextualElement) {
 };
 
 prototype.parseHTML = function(html, contextualElement) {
-  var isSVGContent = (
-    isSVG(this.namespace) &&
-    !svgHTMLIntegrationPoints[contextualElement.tagName]
-  );
-
-  if (isSVGContent) {
+  if (interiorNamespace(contextualElement) === svgNamespace) {
     return buildSVGDOM(html, this);
   } else {
     var nodes = buildHTMLDOM(html, contextualElement, this);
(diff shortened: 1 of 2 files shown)


oneeman:
Whoops, failed tests on other browsers... will fix and update.

oneeman:
ok, done.  I was using a non-existent return value from a function, but didn't realize it because it was passing anyway under chrome and I hadn't tried other browsers before submitting the PR.  Now it's ok.

oneeman:
Also, if you think it'd be better, I could replace these tests with an appropriate `#parseHTML` test in `dom-helper-test.js`.  I had these to help me replicate and pinpoint the bug, but I'm not sure this kind of test is the best here.

packages/morph/lib/dom-helper.js
```diff
@@ -307,6 +307,7 @@ prototype.appendMorph = function(element, contextualElement) {
 };
 
 prototype.parseHTML = function(html, contextualElement) {
+  this.detectNamespace(contextualElement);
   var isSVGContent = (
     isSVG(this.namespace) &&
```
oneeman: Sure, that'd work.  We could even just have something like 
```
  if (interiorNamespace(contextualElement)) {
    return buildSVGDOM(html, this);
  } else {
    var nodes = buildHTMLDOM(html, contextualElement, this);
```
since `interiorNamespace` is doing the same check anyway.  Then `isSVG` would be unused and could be deleted.  Or we could write `isSVG(interiorNamespace(contextualElement)` if we want to be more explicit.

oneeman:
ok, this is ready and passing except for the IE 10 & 11 failures from before.

mixonic:
@oneeman thanks much :cake:

PR #238, opened 2015-01-06
Do not use Array.prototype.map.

rwjblue:
This causes a number of failures on IE8 in Ember.


rwjblue:
Updated to add a small `map` utility function.

mixonic:
Thanks sir :tophat:

PR #239, opened 2015-01-06
Fix <input disabled>

linstula:
Demo showing the same is parsed as `disabled: true` in normal HTML: http://jsbin.com/magegorene/1/

paired with @rwjblue on this.

### DIFF
--- a/packages/htmlbars-compiler/lib/fragment-javascript-compiler.js
+++ b/packages/htmlbars-compiler/lib/fragment-javascript-compiler.js
@@ -61,7 +61,11 @@ FragmentJavaScriptCompiler.prototype.returnNode = function() {
 
 FragmentJavaScriptCompiler.prototype.setAttribute = function(name, value, namespace) {
   var el = 'el'+this.depth;
-  this.source.push(this.indent+'  dom.setProperty('+el+','+string(name)+','+string(value)+(namespace? ',"'+namespace+'"' : '')+');\n');
+  if (namespace) {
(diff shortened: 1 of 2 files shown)


mixonic:
@rwjblue is going to take a quick look here, I'll followup regardless this week.

linstula:
As discussed between @rwjblue and @mixonic we need to change [packages/htmlbars-compiler/lib/fragment-javascript-compiler.js#L64](https://github.com/tildeio/htmlbars/blob/5885113119d28b756baef2a1a185ed51d76f8aa0/packages/htmlbars-compiler/lib/fragment-javascript-compiler.js#L64) to call `setAttribute` instead of `setProperty` to fix this issue. Looks like the changes from #235 made this somewhat more complicated.

oneeman:
ok, this works, go ahead and add it please:
```
FragmentJavaScriptCompiler.prototype.setAttribute = function(name, value, namespace) {
  var el = 'el'+this.depth;
  if (namespace) {
    this.source.push(this.indent+'  dom.setAttributeNS('+el+','+string(namespace)+','+string(name)+','+string(value)+
                     ');\n');
  } else {
    this.source.push(this.indent+'  dom.setAttribute('+el+','+string(name)+','+string(value)+');\n');
  }
};
```

There's a bit of duplication between the `then` and `else` clauses, but trying to eliminate would probably only make it more complicated.

linstula:
@oneeman thanks for putting in the work for this. Should be good to go.

oneeman:
Thank you!  The test failures are from before on IE, so I'll go ahead and merge.

zeppelin:
:+1:

PR #221, opened 2014-12-18
Extracted sauce into an addon / run browsers in parallel

johanneswuerbach:
Extracted sauce into an addon / run browsers in parallel

PR #240, opened 2015-01-07
Improved sauce testing

### LINKED ISSUE #220: Investigate why test failures are not reported
Currently IE 10 and 11 don't pass the htmlbars test suite, but are only marked as finished and not as failed. Also the test output is not published as expected.

Example: https://saucelabs.com/tests/2e3b919384ea4ce48261a61447a2935c

Workaround to get the test output:
1. Open the build, e.g. https://saucelabs.com/tests/2e3b919384ea4ce48261a61447a2935c
2. Open the "Commands" tab
3. Click on "POST execute" and copy the JSON behind "=> " into http://jsonformatter.curiousconcept.com/

johanneswuerbach:
Single tunnel per session
Updated faster tunnel

Extracted from: https://github.com/tildeio/htmlbars/pull/221
Fix #220

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 
 npm-debug.log
 sauce-example.log
+
+testem.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: node_js
 node_js:
 - '0.10'
 sudo: false
-addons:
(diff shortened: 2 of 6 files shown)


johanneswuerbach:
Updated.

PR #242, opened 2015-01-07
Introduce a childAtIndex abstraction

tomdale:
Previously, child nodes were being looked up directly via an element’s
`childNodes` array. However, @krisselden had some concerns around
performance and our future ability to do things like diffing if we have
to precisely simulate the behavior of `childNodes`.

This commit introduces a new `childAtIndex` abstraction that allows us
to create an alternate implementation that walks a linked list instead
of assuming an array implementation.

### DIFF
--- a/packages/morph/lib/dom-helper.js
+++ b/packages/morph/lib/dom-helper.js
@@ -144,6 +144,16 @@ prototype.childAt = function(element, indices) {
   return child;
 };
 
+prototype.childAtIndex = function(element, index) {
+  var child = element.firstChild;
+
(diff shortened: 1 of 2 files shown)


mmun:
@johanneswuerbach any idea what's up?

rwjblue:
Likely related to https://twitter.com/sauceops/status/552919039744282624 (approximately the same time as this PR).  They gave the "all clear" about 22 minutes ago https://twitter.com/sauceops/status/552963243707498498.

I restarted the build.

PR #243, opened 2015-01-08
Report unfinished test runs correctly

johanneswuerbach:


rwjblue:
What is this addressing?

johanneswuerbach:
It looks like IE 8 never actually finished the tests (some js error on the way), but was reported as passing. I'm currently fixing that all browsers are marked as failed....

rwjblue:
Gotcha, awesome!

johanneswuerbach:
Fixed and moved IE 8 into the needs fixing section and IE 9 out.

oneeman:
Thanks!  Should we also move IE 10 and 11 out of the allowed failure group?  They'll be passing after https://github.com/tildeio/htmlbars/pull/244 is merged, and hopefully we can keep them passing in the future.  IE 8 still needs work, though, like you say.

johanneswuerbach:
@oneeman make sense to me

oneeman:
Thanks, merged it now so we're back to a green build. @mmun will remove IE 10 and 11 from allowed failures since they're passing now too.

mmun:
Done in #245.

PR #244, opened 2015-01-08
Call `String` on value in `setAttributeNS` (for IE 10-11)

oneeman:
We may want to handle `null` differently in the future as @mixonic has been considering, but for now this should eliminate the distracting failures in IE 10 and 11.  Not sure about IE 8, it probably has its own issues.

### DIFF
--- a/packages/morph/lib/dom-helper.js
+++ b/packages/morph/lib/dom-helper.js
@@ -149,7 +149,7 @@ prototype.setAttribute = function(element, name, value) {
 };
 
 prototype.setAttributeNS = function(element, namespace, name, value) {
-  element.setAttributeNS(namespace, name, value);
+  element.setAttributeNS(namespace, name, String(value));
 };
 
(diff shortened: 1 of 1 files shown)


oneeman:
IE 8 test got stuck and was killed eventually.  So maybe hold off on merging and I'll try try out an IE 8 VM to see if I can see what's going on, if that's a simple fix too we can do update this PR for that as well.

oneeman:
So I ran the tests with IE8, and it looks like a nightmare.  Besides some relatively simple stuff like the lack of `textContent` (used by several tests), there are many other failures, including some involving SVG, and some which die saying `Died on test #1 undefined: Object doesn't support this property or method` without even giving me the line number as far as I can tell.  I don't think it's something I'm quite up to, so this PR should just be as is for IE 10 & 11.

rwjblue:
Restarted travis build.

PR #247, opened 2015-01-10
IE8 compatibility

salzhrani:
First batch of changes ... feedback is greatly appreciated

### DIFF
--- a/packages/htmlbars-compiler/tests/fragment-test.js
+++ b/packages/htmlbars-compiler/tests/fragment-test.js
@@ -5,7 +5,7 @@ import HydrationJavaScriptCompiler from "../htmlbars-compiler/hydration-javascri
 import { DOMHelper } from "../morph";
 import { preprocess } from "../htmlbars-syntax/parser";
 import { get } from "../htmlbars-runtime/hooks";
-import { equalHTML } from "../htmlbars-test-helpers";
+import { equalHTML, getTextContent } from "../htmlbars-test-helpers";
 
 var xhtmlNamespace = "http://www.w3.org/1999/xhtml",
     svgNamespace = "http://www.w3.org/2000/svg";
@@ -48,20 +48,22 @@ test('compiles a fragment', function () {
   equalHTML(fragment, "<div> bar </div>");
 });
 
-test('compiles an svg fragment', function () {
-  var ast = preprocess("<div><svg><circle/><foreignObject><span></span></foreignObject></svg></div>");
-  var fragment = fragmentFor(ast);
+if (document && document.createElementNS) {
+  test('compiles an svg fragment', function () {
+    var ast = preprocess("<div><svg><circle/><foreignObject><span></span></foreignObject></svg></div>");
+    var fragment = fragmentFor(ast);
 
-  equal( fragment.childNodes[0].namespaceURI, svgNamespace,
-         'svg has the right namespace' );
-  equal( fragment.childNodes[0].childNodes[0].namespaceURI, svgNamespace,
-         'circle has the right namespace' );
-  equal( fragment.childNodes[0].childNodes[1].namespaceURI, svgNamespace,
-         'foreignObject has the right namespace' );
-  equal( fragment.childNodes[0].childNodes[1].childNodes[0].namespaceURI, xhtmlNamespace,
-         'span has the right namespace' );
-});
-
+    equal( fragment.childNodes[0].namespaceURI, svgNamespace,
+           'svg has the right namespace' );
+    equal( fragment.childNodes[0].childNodes[0].namespaceURI, svgNamespace,
+           'circle has the right namespace' );
+    equal( fragment.childNodes[0].childNodes[1].namespaceURI, svgNamespace,
+           'foreignObject has the right namespace' );
+    equal( fragment.childNodes[0].childNodes[1].childNodes[0].namespaceURI, xhtmlNamespace,
+           'span has the right namespace' );
+  });
+}
+  
 test('compiles an svg element with classes', function () {
   var ast = preprocess('<svg class="red right hand"></svg>');
   var fragment = fragmentFor(ast);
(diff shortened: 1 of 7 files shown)


oneeman:
Awesome, thanks for tackling this!  It'd be great to have IE8 passing.  Here's what I've got:

* For skipping tests involving svg, if possible place those tests all together so they can be skipped with just one such conditional per file.  And use the same test condition for consistency (unless there's a reason not to which I'm not aware of).  What I mean is in some you check whether `createElementNS` exists and in others whether a `div` has `namespaceURI`, instead use just one of those two if possible.  Perhaps even move that feature detection to the test helper and give its result a descriptive name, then use that boolean in the files with tests.
* Also for the skipping, I noticed the style used so far has been to not indent the contents of the `if` for skipping at all.  I assume we want to keep it that way.
* For the fixes for `textContent`,  don't use the ternary operator or `if`'s  since it just makes less clear than it needs to be.  Instead, if the one that works for IE8 works for other browsers, just use that.

Other comments will be as line comments.

packages/htmlbars-syntax/lib/utils.js
```diff
@@ -17,7 +17,14 @@ export function parseComponentBlockParams(element, program) {
...
 
-  var asIndex = attrNames.indexOf('as');
+  var asIndex = -1;
+
+  for (i = 0; i < attrNames.length; i++) {
+    if (attrNames[i] === 'as') {
+      asIndex = i;
+      break;
+    }
+  }
```
oneeman: Wow, I had no idea.  For the `for` loop, please use `i < l` rather than `i < attrNames.length`.  It's customary to do this for performance reasons, and you often see stuff like `for (var i=0, l=items.length; i < l; i++)` in the ember codebase.  In this case, we already have `l` from above.

packages/htmlbars-compiler/tests/html-compiler-test.js
```diff
@@ -199,12 +212,22 @@ function shouldBeVoid(tagName) {
...
+  var closing;
+  if (div.innerHTML.indexOf(':') >= 0) {
+    tag = '<:' + tagName + ' data-foo="bar">';
+    closing = '</:' + tagName + '>';
+  } else {
+    tag = '<' + tagName + ' data-foo="bar">';
+    closing = '</' + tagName + '>';
+  }
+    
   var extra = "<p>hello</p>";
```
oneeman: Please implement the fix for this problem in `normalizeInnerHTML` in the `ie8InnerHTML` clause.  Should be doable with a regex search like the other fixes there.

oneeman:
Also, could you please remove the `allow_failures` section in `.travis.yml`?  IE8 is the last one failing, so we shouldn't need that section anymore after this PR.

packages/htmlbars-syntax/lib/utils.js
mixonic: @salzhrani I encourage you to create an `indexOfArray` method that is the native version if available, and a poly-fill if not.

I suggest reviewing [ember's implementation](https://github.com/emberjs/ember.js/blob/d677cbb0c51eee34910ab4075bc53c938d0e3055/packages/ember-metal/lib/array.js#L62) for inspiration!

packages/morph/lib/dom-helper.js
```diff
@@ -141,7 +141,7 @@ prototype.childAt = function(element, indices) {
 };
 
 prototype.childAtIndex = function(element, index) {
-  return element.childNodes.item(index);
+  return element.childNodes.length > index ? element.childNodes.item(index) : null;
```
mixonic: Why is this needed?

packages/morph/lib/dom-helper.js
```diff
@@ -153,7 +153,11 @@ prototype.setAttribute = function(element, name, value) {
 };
 
 prototype.setAttributeNS = function(element, namespace, name, value) {
-  element.setAttributeNS(namespace, name, String(value));
+  if (element.setAttributeNS) {
```
mixonic: Is this feature detection need on an element-by-element basis, or can we treat it as a browser feature detect?

```js
// I'm sure this detect can be written better, see the top of the file for others
if (document && document.createElement('div').setAttributeNS)  {
  prototype.setAttributeNS = function(element, namespace, name, value) {
    element.setAttributeNS(namespace, name, String(value));
  }
} else {
  prototype.setAttributeNS = function(element, namespace, name, value) {
    element.setAttribute(name, String(value));
  }
}
```

packages/htmlbars-compiler/tests/fragment-test.js
```diff
@@ -178,13 +183,13 @@ test('test auto insertion of text nodes for needed edges a fragment with morph m
...
   equal(t.nodeType, 3);
-  equal(t.textContent , '');
+  equal(t.textContent !== undefined ? t.textContent : t.toString() , '');
   equal(morphs[1].start, null);
   equal(morphs[1].end, null);
 
   equal(morphs[2].start, morphs[1].parent());
   equal(morphs[2].end.nodeType, 3);
-  equal(morphs[2].end.textContent, '');
+  equal(morphs[2].end.textContent  !== undefined ? morphs[2].end.textContent : morphs[2].end.toString(), '');
```
mixonic: These assertions with ternary statements are really confusing. They should be the same for all browsers.

mixonic:
Please do group the SVG tests together. I generally prefer to not indent in the test files because it keeps diffs much easier to read.

mixonic:
@salzhrani if you have time this week, we will need to get IE8 fixes in before the weekend. It would be a huge help if we could just merge this PR!

salzhrani:
I understand and I am trying my best. I have time on Friday and I believe I can get it done by then.

oneeman:
Much appreciated!  Since I've got time this week, how about I work on the remaining failures (which don't look too bad... at least they're failing rather than dying like the ones you debugged were!), and when you have time you can finalize what you already have here?  Then, between the two, IE8 should be green.

salzhrani:
I'll give be it a swing tonight to incorporate the feedback and fix a few more issues and @oneeman can take it from there.

oneeman:
Awesome.  Please rebase and squash as well as there have been a lot of commits to master the last few days.  I made a bit of progress last night, which you can check out at: https://github.com/oneeman/htmlbars/commits/more-ie8, so as to avoid duplicating effort when you work on remaining issues.  Once your changes are done I can rebase mine as well and keep going.  And I'll be on IRC tonight (CST).

packages/morph/lib/dom-helper.js
oneeman: This should take care of https://github.com/tildeio/htmlbars/issues/259

salzhrani:
rebased and squashed. Failed tests are down to 27, Thank you all for the feedback. @oneeman appreciate you tackling the remaining issues.

packages/morph/tests/dom-helper-test.js
```diff
@@ -56,9 +57,9 @@ test('#setAttribute', function(){
 
   node = dom.createElement('input');
   dom.setAttribute(node, 'disabled', true);
-  equalHTML(node, '<input disabled="true">');
+  equalHTML(node, ie8 ? '<input disabled>' : '<input disabled="true">');
```
mixonic: This can just assert `ok(!!node.getAttribute('disabled'))`.

Then we can avoid the export of `ie8` (we don't detect browsers anywhere else, just features) and use of it only in these tests.

salzhrani:
This is as far as I can take this PR given the required expertise and time constraints. Thank you all

oneeman:
Absolutely, thank you!  I noticed you managed to do some additional clean-up and even fix some more failing tests.  We can merge and I'll continue on top of this in another PR.

mmun:
Great work! :+1:

mixonic:
@salzhrani thanks for helping to shoulder the burden of IE8 with us :-) I appreciate your patience with my comments.

salzhrani:
It is a pleasure. I look forward to more comments .

PR #248, opened 2015-01-10
Change setProperty and morph to remove an undefined or NaN attr value

mixonic:
Open for debate. Maybe this should be part of another layer in Ember. IMO it is ok here.

Ember expects setting an attribute binding to `undefined` or `NaN` to remove the attribute entirely. This follows that logic.

### DIFF
--- a/packages/morph/lib/attr-morph.js
+++ b/packages/morph/lib/attr-morph.js
@@ -1,5 +1,5 @@
 import { sanitizeAttributeValue } from "./attr-morph/sanitize-attribute-value";
-import { normalizeProperty } from "./dom-helper/prop";
+import { isAttrRemovalValue, normalizeProperty } from "./dom-helper/prop";
 import { svgNamespace } from "./dom-helper/build-html-dom";
 
 function updateProperty(value) {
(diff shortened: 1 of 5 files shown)


oneeman:
Does `updateAttributeNS` also need to be updated?  And this might be a good opportunity to DRY it up a bit as well (e.g. L170-178 and L184-192 in `dom-helper.js`, but maybe also the check for remove attribute between there and `attr-morph`).

As for whether it belongs here, maybe someone else can chime in.  I don't know the ember code yet, so I wouldn't know what should go where.

mixonic:
The Ember side is here: https://github.com/emberjs/ember.js/pull/10186

mixonic:
Phantom has no native `isNaN`. `function isNan(x) { return x !== x }` is a possible poly-fill.

mixonic:
After discussion with @rwjblue I've dropped the `NaN` check.

In Ember, I'm leaving the `NaN` value behavior undefined: https://github.com/emberjs/ember.js/pull/10186/files#diff-f7920bfff32661282a5d9c883cc26641L43

oneeman:
Thanks!

rwjblue:
@mixonic - Let me know when you need a version cut for the Ember updates.

PR #250, opened 2015-01-10
[FEATURE] Allow createElement to take a hash of attributes

ymainier:


oneeman:
Thanks!  I'm not really familiar with the issue, so I can't comment on the generals, but I did spot a problem with the attribute namespaces.  You're setting their namespace according to that of the element (i.e. attributes on svg elements are getting svg namespace).  What should happen is that the attributes either have no namespace or are assigned the correct namespace based on the prefix in the attribute name (e.g. `xlink:href` needs the `xlink` namespace).  The way we're doing it is that the namespace for an attribute is passed in as an argument to the various methods (`setAttributeNS`, `setProperty`, and the ones for the morphs).  So, something similar would need to be done here (e.g. the attributes in the hash passed in to `createElement` would contain a `namespace` property which is what you'd use to know whether the attribute needs a namespace or not and if so which).

For some more details on this, see https://github.com/tildeio/htmlbars/pull/235 and https://github.com/tildeio/htmlbars/issues/224

ymainier:
Thanks for your clarification !
I have two questions : 
1. Can setAttributeNS be used on an element that was not created by createElementNS? I've just read http://www.w3.org/TR/DOM-Level-2-Core/core.html#ID-ElSetAttrNS and it is still not clear to me.
2. Regarding the namespace property in the namespace hash, what do you think of the following example?
```javascript
{
  "xlink:href" : {
    value : "cool-script.js",
    namespace : ""
  },
  type : {
    value : "text/ecmascript"
  }
}
```

oneeman:
Let me punt to @mixonic on this, as he has experience with SVG and this kind of attribute, as well as the goals of the feature.

PR #252, opened 2015-01-12
Add namespace detection in `AttrMorph`

oneeman:
This will allow `bind-attr` to work with namespaced attributes (no further changes needed), as well as eventually attributes bound via ember views (so long as the full attribute name is passed to `createAttrMorph`).  Also added a test for the bound case, which wasn't covered.

### DIFF
--- a/packages/htmlbars-util/lib/namespaces.js
+++ b/packages/htmlbars-util/lib/namespaces.js
@@ -1,0 +1,20 @@
+// ref http://dev.w3.org/html5/spec-LC/namespaces.html
+var defaultNamespaces = {
+  html: 'http://www.w3.org/1999/xhtml',
+  mathml: 'http://www.w3.org/1998/Math/MathML',
+  svg: 'http://www.w3.org/2000/svg',
+  xlink: 'http://www.w3.org/1999/xlink',
+  xml: 'http://www.w3.org/XML/1998/namespace'
+};
+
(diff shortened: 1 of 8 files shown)


oneeman:
Updated to only detect the namespace in `AttrMorph` if it's not received as an argument.

packages/htmlbars-util/lib/namespaces.js
```diff
@@ -0,0 +1,19 @@
...
+var defaultNamespaces = {
+  html: 'http://www.w3.org/1999/xhtml',
+  mathml: 'http://www.w3.org/1998/Math/MathML',
+  svg: 'http://www.w3.org/2000/svg',
+  xlink: 'http://www.w3.org/1999/xlink',
+  xml: 'http://www.w3.org/XML/1998/namespace'
+};
+
+export function getAttrNamespace(attrName) {
+  var parts = attrName.split(':');
```
mmun: Can you do an indexOf and check slices instead of always splitting?

oneeman: Done :-)  Apologies for the lame variable name `colonIndex`.

mmun:
LGTM :+1:

mixonic:
Awesome! Thanks @oneeman!

rwjblue:
Will release shortly.

PR #253, opened 2015-01-12
Allow numbers to be parsed as HTML in IE

mixonic:
I some cases (like select tests in Ember), numbers may be passed to `parseHTML`. This fixes that code path for IE <10.


rwjblue:
LGTM

Issue #254, opened 2015-01-13
Need better error for unclosed element.


rwjblue:
Thrown [here](https://github.com/tildeio/htmlbars/blob/cda0d6d6073e9baea2c580335b630a683286ef69/packages/htmlbars-syntax/lib/node-handlers.js#L30).  It should contain the line number of the unclosed tag.

oneeman:
Nice catch, thanks!

PR #255, opened 2015-01-13
Include line number in error thrown for unclosed html element (#254)

oneeman:
For https://github.com/tildeio/htmlbars/issues/254.

Includes tests for both the basic case and one with two unclosed elements.  The implementation will always throw on the last unclosed one left.  So the second test is currently testing an implementation detail.  I'd be open to removing it or modifying it to only check that some error is thrown but be agnostic as to which element it's about.


rwjblue:
Awesome!

PR #256, opened 2015-01-14
removeAttribute fix for IE <11 and svg

### LINKED ISSUE #251: IE <11 does not honor removeAttribute
Try it with viewBox: http://jsfiddle.net/2BqmC/2/

However setting the value of the attribute to `null` does seem to work at first glance. `domHelper.removeAttribute` needs to handle this browser bug.

mixonic:
Fixes #251

Old IE cannot remove svg attributes. Create a slow path for it.

### DIFF
--- a/packages/morph/lib/dom-helper.js
+++ b/packages/morph/lib/dom-helper.js
@@ -30,6 +30,13 @@ var ignoresCheckedAttribute = doc && (function(document){
   return !clonedElement.checked;
 })(doc);
 
+var canRemoveSvgViewBoxAttribute = doc && (doc.createElementNS ? (function(document){
+  var element = document.createElementNS(svgNamespace, 'svg');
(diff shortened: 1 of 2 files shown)


rwjblue:
@mixonic - Any progress on this?

mixonic:
Need to make a map. Blah. Will wrap it this evening.

mixonic:
This is good to go. Detecting this behavior is really hard, so I suggest we just make a narrow exception and grow the whitelist if needed.

rwjblue:
@mixonic - Sounds good.  Will merge once Travis is happy.

mixonic:
womp womp will fix tonight.

rwjblue:
Travis failure is IE8 only (we recently removed it from the allowed failure list) on this test.

screenshot

PR #258, opened 2015-01-14
Hide passed tests by default.

rwjblue:
This should make the default screenshots by saucelabs more helpful.


oneeman:
I see the difference in the IE8 screencast, thanks!

rwjblue:
Example screenshot from [sauce-labs job](https://saucelabs.com/tests/267105f643e347459a42ecdbfec7c82c):

screenshot

PR #260, opened 2015-01-15
Finish fixes for IE8

oneeman:
A bit of clean-up, and getting all the failing/dying tests green.  Should be ready before too long, but it's never too early for feedback.  I am doing very short amend commits, though, so it might be a good idea to wait a bit until it's closer to being finalized... up to you.

### DIFF
--- a/packages/htmlbars-compiler/tests/html-compiler-test.js
+++ b/packages/htmlbars-compiler/tests/html-compiler-test.js
@@ -5,12 +5,33 @@ import defaultHooks from "../htmlbars-runtime/hooks";
 import defaultHelpers from "../htmlbars-runtime/helpers";
 import { merge } from "../htmlbars-util/object-utils";
 import { DOMHelper } from "../morph";
-import { normalizeInnerHTML, getTextContent } from "../htmlbars-test-helpers";
+import { createObject, normalizeInnerHTML, getTextContent } from "../htmlbars-test-helpers";
 
 var xhtmlNamespace = "http://www.w3.org/1999/xhtml",
     svgNamespace   = "http://www.w3.org/2000/svg";
 
 var hooks, helpers, partials, env;
+
+// IE8 doesn't correctly handle these html strings
+var innerHTMLSupportsCustomTags = (function() {
+  var div = document.createElement('div');
+  div.innerHTML = '<x-bar></x-bar>';
+  return normalizeInnerHTML(div.innerHTML) === '<x-bar></x-bar>';
+})();
+
+// IE8 doesn't handle newlines in innerHTML correctly
+var innerHTMLHandlesNewlines = (function() {
+  var div = document.createElement('div');
+  div.innerHTML = 'foo\n\nbar';
+  return div.innerHTML.length === 8;
+})();
+
+// IE8 removes comments and does other unspeakable things with innerHTML
+var ie8GenerateTokensNeeded = (function() {
+  var div = document.createElement("div");
+  div.innerHTML = "<!-- foobar -->";
+  return div.innerHTML === "";
+})();
 
 function registerHelper(name, callback) {
   helpers[name] = callback;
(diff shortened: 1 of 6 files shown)


packages/htmlbars-compiler/tests/html-compiler-test.js
```diff
@@ -796,6 +785,15 @@ test('Components - Called as helpers', function () {
   compilesTo('a<x-append text="d{{bar}}">b{{baz}}</x-append>f','abcdef', object);
 });
 
+// IE8 doesn't correctly handle these html strings, so in that case we skip them
+var innerHTMLSupportsCustomTags = (function() {
+  var div = document.createElement('div');
+  div.innerHTML = '<x-bar></x-bar>';
+  return normalizeInnerHTML(div.innerHTML) === '<x-bar></x-bar>';
+})();
```
mixonic: Generally nice to keep the feature detects together at the top of the file.

mixonic:
Lookin' good. Will wait until you remove `WIP` from the title and say good-to-go @oneeman.

packages/htmlbars-compiler/tests/html-compiler-test.js
oneeman: done (except in the test-helper, where it basically all feature detections and fixes for them)

oneeman:
This is ready, with some comments:

* The only change to production code is in splitting by newlines in `parser.js`.  Everything else is in the tests
* The test, `#setProperty removes namespaced attr with undefined`, which I extracted and skipped, would sometimes pass and sometimes fail with IE8.  When you ran it by itself it'd always pass, when you ran the entire suite it'd alternate between passing and failing.  What was going on was that it would get a different result from `normalizeProperty` in `setProperty`, and go down a different branch (one of which passes, the other fails).  It seems the tests are not atomic, and the property cache was leaking and leading to this result because QUnit runs the failing tests from the previous round first. So after it fails, the next time it runs first and passes.  This might have broader implications for testing strategies (as might the cache code in the template?), so I thought I'd mention it even though this test is skipped in IE8 now
* I generally grouped the skipped tests together.  One notable exception is `The compiler can handle newlines`, which seemed to really fit in where it was.
* Saucelabs is on the fritz, in several builds (872) it reported different browsers as failing on different builds, but the screenshots for them showed 0 failures (except one safari build where the screenshot was before it was done).  And IE8 hung on one of the builds.  So for now I'm putting the travis configuration back to how it was (which is now build 873).  Doesn't seem to have helped, though, it reports IE8 failure on the current one.  Not sure what to do about it.
* The `innerHTML` workarounds (e.g. in `generateTokens`) are pretty hacky, but I couldn't think of anything else short of working to become an expert on the many intricacies of IE8's insanity.

johanneswuerbach:
@oneeman I'll look into the sauce errors asap. Looks like there is a race condition in https://github.com/igorlima/sauce-js-tests-integration, I guess it finally needs tests ....

oneeman:
@johanneswuerbach Thanks!

oneeman:
Actually, I'm thinking I was wrong to skip the line number tests for the compiler, since those don't only involve the compiler.  So I'll look into that and if possible fix and re-enable them.  But I'll flag any additional changes with line comments here, so the code is ready for review with that caveat.

packages/htmlbars-compiler/tests/html-compiler-test.js
```diff
@@ -1012,6 +1013,8 @@ test("A helpful error message is provided for mismatched start/end tags", functi
   }, /Closing tag `div` \(on line 5\) did not match last open tag `p` \(on line 2\)\./);
 });
 
+if (innerHTMLHandlesNewlines) {
```
oneeman: These skipped tests don't actually have anything to do with `innerHTML`, so this should probably be a separate feature detect (also involving newlines in IE8, but not `innerHTML`), or rename this feature to be more generic.  That is, unless we find a way to get those tests passing so they don't have to be skipped.

mixonic: Sure, an inline comment would be helpful so we don't forget.

If you make that change, please merge! :-)

mixonic:
Thank you, thank you, thank you, thank you, thank you.

packages/htmlbars-compiler/tests/html-compiler-test.js
mixonic: I'll add it in another commit :-) Eager to get this in.

salzhrani:
Awesome work.

PR #262, opened 2015-01-16
Improved test run finished detection

johanneswuerbach:
Before some async tests could finish after the result was published.


johanneswuerbach:
IE 8 passed two times in a row so I removed all allowed failures. :joy: 

Fixes: https://github.com/tildeio/htmlbars/pull/260#issuecomment-70201399

oneeman:
lol, I was just about to ask if you could remove the allowed failure.  Thanks a ton!

rwjblue:
ZOMG

PR #264, opened 2015-01-19
Use linked list walking for childAtIndex

tomdale:
This implementation entirely removes the dependency on an element’s
`childNodes` live collection. This is beneficial because:

1. IE8 required a bounds check that relied on childNodes.length, which
   in turn meant we had to implement an array-backed live collection of childNodes in simple-dom
   which is slow and error-prone, and
3. The browser just uses a linked list anyway and we were forcing *it*
   to materialize and keep updated a live collection until GC time.

This implementation just does what the browser would do anyway, while
making our simple-dom implementation actually, well, simple, and
includes notes about all of this for future fearless adventurers who
find themselves in this area of the code.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -5,6 +5,7 @@
     "test": true,
     "expect": true,
     "equal": true,
+    "strictEqual": true,
     "deepEqual": true,
     "ok": true
   },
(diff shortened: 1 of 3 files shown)


rwjblue:
LGTM

PR #267, opened 2015-01-23
Update ARCHITECTURE.md

eccegordo:
fix broken links to parser.js
[parser.js](packages/htmlbars-syntax/lib/parser.js)

ast.js link also seems to be broken, but not sure where it is suppose to go.


rwjblue:
Can you squash the commits?

PR #273, opened 2015-01-30
Add hook for extracting protocol from URLs

tomdale:
In order to sanitize potentially dangerous URLs that contain executable
behavior (e.g. “javascript:” URLs), we need to determine the protocol.

Unfortunately, URL parsing is notoriously error-prone, so we want to use
the host environment’s native functionality such that the protocol we
report is the same as what it will act upon.

In this case, we expose a `protocolForURL` hook that uses a generated
`<a>` element to set its `href` and check the resulting `protocol`. A
Node.js implementation could fall back to using the `url` package that
is included in the standard library.

### DIFF
--- a/packages/morph/lib/dom-helper.js
+++ b/packages/morph/lib/dom-helper.js
@@ -403,4 +403,16 @@ prototype.parseHTML = function(html, contextualElement) {
   }
 };
 
+var parsingNode;
+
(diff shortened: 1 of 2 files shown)


mmun:
:+1:

rwjblue:
Will release shortly.

PR #277, opened 2015-02-01
WIP Add a failing test demonstrating IE8 space shifting

mixonic:
This test highlights a bug in IE8 shifting spaces onto previous text nodes. It fails as follows:

The `"ohhi "` text node is given the space text node, making it `"ohhi  "`. (*sigh* and github makes the two spaces after the second `ohhi` here a single space).

For the clone node repairing to handle this, it needs to support splitting the existing text nodes as well as injecting fresh blank text nodes.

An alternative solution would be not cloning the DOM in IE8 at all.

### DIFF
--- a/packages/dom-helper/tests/dom-helper-test.js
+++ b/packages/dom-helper/tests/dom-helper-test.js
@@ -404,7 +404,7 @@ test('#cloneNode deep', function(){
   equalHTML(node, '<div><span></span></div>');
 });
 
-test('dom node has empty text after cloning and ensuringBlankTextNode', function(){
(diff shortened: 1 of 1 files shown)


rwjblue:
I'm :+1: for disabling clone node in IE8.

mixonic:
After some discussion, it sounds like the ideal plan is to **disabled cloning the fragment on < IE9**. I believe this does not mean the repair code can be completely removed, but it can probably be sliced down.

The endgame here is getting Ember tests themselves to pass in IE8.

mixonic:
#278 addressed this.

PR #278, opened 2015-02-01
Only use `cloneNode` when appropriate

oneeman:
Will implement what's discussed in https://github.com/tildeio/htmlbars/pull/277.  However, I'm stuck without an IE8 VM at the moment, so am using saucelabs as my IE8 testing rig.  So I'm opening this PR before it's ready.  Will remove the WIP from the title and let you know when it's done.


oneeman:
ok, this is ready.  It was failing with IE8 before and passes now, and I verified that with chrome it does use `cloneNode` (i.e. that the feature detection doesn't give a false negative).

Because of the `env.useCacheFragment` condition, this will require changes in Ember to along with it, in order to use cloning normally.

mixonic:
Thanks @oneeman :-)

PR #279, opened 2015-02-02
DOMHelper's parseHTML now returns a document fragment

mmun:
Previously it would return a NodeList, now it returns a document fragment.


mmun:
Hmm... seems to be failing on some browsers.

rwjblue:
I believe that this is a transient IE8 failure, restarting.

mmun:
I'm curious to benchmark this code and try rewriting it without touching `childNodes`.

PR #282, opened 2015-02-03
[WIP] Reactive rerendering

wycats:
This is a WIP PR to implement React-inspired fast rerendering of HTMLBars templates.

One important improvement we can make over the React model is eliminating any work for static portions of the template. Because HTMLBars templates are declarative, not imperative like React's model, the compiled template knows which parts of the DOM can change (because they were created through `{{}}`s) and which parts will never change.

Compiled HTMLBars templates already separate out the process of building the static "skeleton" from the process of filling it in, and this PR allows us to update the dynamic portions of the result of a previous render. Conceptually, this allows us to run the same "rendering" code multiple times, making re-renders fast without having to expensively diff parts of the DOM that will never change.

Because conditionals and loops are **also** declarative, we can eliminate the cognitive overhead React requires when conditionally rendering content. Instead of requiring that the user [remember to insert `null` into a conditional slot](https://facebook.github.io/react/blog/2014/07/13/react-v0.11-rc1.html#rendering-to-null), the template engine will know the position of an `{{#if}}` block, regardless of whether the conditional is truthy or not. There is no way for the user to screw this up and produce an off-by-one misalignment, which circumvents the diffing algorithm and triggers a full re-render.

#### TODO

- [ ] ~~Make inline helpers render nodes~~
- [x] Provide render nodes to inline helpers
- [ ] Write a test for (and implement) re-rendering `{{yield}}` for components
- [x] Support simpler pass-thru rendering in block helpers (`options.render`)
- [ ] Investigate issue where first or last morph is shifted out of the way and has child morphs
- [x] Propagate `rootRenderNode` through the render node hierarchy
- [x] Implement infrastructure to support render node dirtying
- [ ] Implement general infrastructure for working with framework-provided "opaque values"
- [ ] Related to the previous bullet: generally make it easier to "subclass" the hooks, by separating "derived" hooks that delegate to "fundamental" hooks that must be overridden.
- [ ] Move shared helper functions into `utils`


wycats:
The implementation in this PR is **extremely** WIP, so don't let the implementation scare you :wink:

mkoxio:
looks amazing!

cibernox:
One question about how the diff-ing works. 
If an template contains a triple moustache {{{}}} where I output trusted HTML and I re-render that template, the content of that moustache will be diffed? 

That would be just beautiful.

mixonic:
@cibernox Others can correct me if I am wrong, but this work does not make any effort toward that aim.

In fact, most libraries the perform dom-diffing are performing dom-like-data-tree diffing. The "virtual DOM" as we call it. They don't actually parse HTML into DOM, they require you to use a DSL to write dom-like objects that are later translated into DOM.

Since a value inside a `{{{}}}` is an HTML string, the best way to parse it into DOM is going to be using `innerHTML`. This returns real DOM objects, we any implementation wouldn't be diffing a virtual dom-like data structure but actual DOM. There might be performance gains, but it is definitely unlike any diffing solution out there as I understand them.

Additionally, the case of `{{{}}}` is extremely rare. Ember applications use templates heavily, and it is there we can do the work of determining structure upfront (much as virtual-dom DSLs do) and make an incredible impact.

Hope that makes some sense!

stefanpenner:
> Since a value inside a {{{}}} is an HTML string, the best way to parse it into DOM is going to be using innerHTML. This returns real DOM objects, we any implementation wouldn't be diffing a virtual dom-like data structure but actual DOM. There might be performance gains, but it is definitely unlike any diffing solution out there as I understand them.

or just old{{{}}} === new{{{}}} :P

cibernox:
I imagined that. 

I had a very particular use-case where some cards contain safe HTML coming directly from a CMS. Although I can imagine a workaround. Converting the HTML to templates in runtime (I would have to include the HTMLBars compiler) and render views dinamically instead of using {{{}}}.

Sounds possible in my head :P

packages/htmlbars-compiler/lib/compiler.js
```diff
@@ -1,6 +1,7 @@
 /*jshint evil:true*/
 import { preprocess } from "../htmlbars-syntax/parser";
 import TemplateCompiler from "./template-compiler";
+import { wrap } from "../htmlbars-runtime/hooks";
```
rwjblue: shouldn't be importing from `htmlbars-runtime` from within `htmlbars-compiler` for anything other than tests (the runtime is not available within Ember for example).  If the wrap function needs to be shared and available to compiler (and ultimately to the consuming library) it should be in `htmlbars-utils`.

wycats: Nice. If `utils` is imported into both `runtime` and `compiler`, that's the right place for this for sure.

mmun:
Continued in #318.

PR #283, opened 2015-02-03
compilerSpec -> compileSpec

Dremora:
I've tried including `htmlbars-compiler` into my project and discovered that `compilerSpec` is not defined. Instead, `htmlbars-compiler/compiler` exports a function named `compileSpec`, which `htmlbars-compiler` mistakingly imports and re-exports as `compilerSpec`. It seems this file is not used anywhere else neither within HTMLBars nor in Ember. That's the only occurrence of the string `compilerSpec` in the entire project.


mmun:
Thank you!

PR #289, opened 2015-02-06
Only use `cloneNode` when appropriate

rwjblue:
- Add a previously failing test for IE8
- Add `env.useFragmentCache` and `dom.canClone`
- Update templates to only use `cloneNode` when both those are true

---

Cherry pick f9d97f06ee60d875275b630d0c338eaf83dbb689 from https://github.com/tildeio/htmlbars/pull/278 into the 0-8-stable branch.


rwjblue:
Submitted this here to get Travis to confirm against all browsers.  Green, so merging to release a 0.8.5.

Issue #290, opened 2015-02-06
Support latest Handlebars AST


mmun:
Statement nodes explicitly list their parts (e.g. name, params, hash, etc.) instead of using a subexpression. See the discussion [here](https://github.com/wycats/handlebars.js/pull/941) in particular [this documentation update](https://github.com/wycats/handlebars.js/commit/999da739a66199483ffc4d82426550aee5ac798f?diff=split) is enlightening.

Anyone interested? I can break it down further.

mmun:
Completed in #291.

PR #291, opened 2015-02-09
Update handlebars AST

marcioj:
As requested by @mmun in #290

### DIFF
--- a/packages/htmlbars-compiler/lib/hydration-opcode-compiler.js
+++ b/packages/htmlbars-compiler/lib/hydration-opcode-compiler.js
@@ -121,7 +121,9 @@ HydrationOpcodeCompiler.prototype.mustache = function(mustache, childIndex, chil
   this.morphs.push([morphNum, this.paths.slice(), start, end, mustache.escaped]);
 
   if (isHelper(mustache)) {
-    prepareCommonOpcodes(this, mustache);
+    prepareHash(this, mustache.hash);
+    prepareParams(this, mustache.params);
+    preparePath(this, mustache.path);
     this.opcode('printInlineHook', morphNum);
   } else {
     preparePath(this, mustache.path);
@@ -140,7 +142,9 @@ HydrationOpcodeCompiler.prototype.block = function(block, childIndex, childCount
   var templateId = this.templateId++;
   var inverseId = block.inverse === null ? null : this.templateId++;
 
-  prepareCommonOpcodes(this, block);
+  prepareHash(this, block.hash);
+  prepareParams(this, block.params);
+  preparePath(this, block.path);
   this.opcode('printBlockHook', morphNum, templateId, inverseId);
 };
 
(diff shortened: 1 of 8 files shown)


mmun:
You should rename the `isHelper` param from sexpr to mustache. Sexprs are always helpers now.

packages/htmlbars-compiler/lib/hydration-opcode-compiler.js
```diff
@@ -121,19 +114,17 @@ HydrationOpcodeCompiler.prototype.closeElement = function() {
...
 
   var morphNum = this.morphNum++;
   var start = this.currentDOMChildIndex;
   var end = this.currentDOMChildIndex;
   this.morphs.push([morphNum, this.paths.slice(), start, end, mustache.escaped]);
 
-  if (isHelper(sexpr)) {
-    prepareSexpr(this, sexpr);
+  if (isHelper(mustache)) {
+    prepareCommonOpcodes(this, mustache);
```
mmun: I'd prefer to just expand this out to the 3 calls so that it stays at the same abstraction as the AST.

packages/htmlbars-compiler/lib/hydration-opcode-compiler.js
```diff
@@ -233,8 +222,13 @@ HydrationOpcodeCompiler.prototype.pushMorphPlaceholderNode = function() {
   this.comment();
 };
 
+HydrationOpcodeCompiler.prototype.MustacheStatement = function(mustache) {
```
mmun: Why is this here? Seems strange to have both this AND `HydrationOpcodeCompiler.prototype.mustache`, no?

packages/htmlbars-syntax/tests/parser-node-test.js
```diff
@@ -286,42 +286,42 @@ test("Node helpers", function() {
 test("Tokenizer: MustacheStatement encountered in tagName state", function() {
   var t = "<input{{bar}}>";
   astEqual(t, b.program([
-    b.element('input', [], [ b.sexpr(b.path('bar')) ])
+    b.element('input', [], [ b.mustache(b.path('bar')) ])
```
mmun: Hm, these nodes really want to be their own AST node. Something like `ElementModifier`.

packages/htmlbars-compiler/lib/hydration-opcode-compiler.js
marcioj: This method runs in the context of `ConcatStatement` like `<div foo="something {{bar}}">`. Due to AST update, now it is generated `concat(text("something"), mustache("bar"))` instead of  `concat(text("something"), sexp("bar"))`, so `MustacheStatement` is visited. 
I don't have a deep understanding of what is happening, but I believe that mustaches in attributes behaves different from these outside (Morphs vs AttrMorphs), so using the mustache method in that case could generate a different behavior.

mmun:
:+1: I'm going to defer merging this until the reactive rerending stuff is merged.

packages/htmlbars-syntax/tests/parser-node-test.js
marcioj: Sorry, not sure if I get the idea. Do you want to group the `tag`, `attributes` and `helpers` property of `ElementNode` in a node called `ElementModifier`?

mmun: In another PR I want to rename `ElementNode.helpers` &rarr; `ElementNode.modifiers` and change them from `MustacheStatement` to `ElementModifier`.

mmun:
I'm going to tweak your PR and merge it soon, now that Handlebars 3.0.0 has been released.

mmun:
Merged manually in 4f089543c470c4a73f9b068a2dca867d8ce15c53.

marcioj:
Glad to help!

PR #294, opened 2015-02-11
Use ember-cli for tests instead of testem directly

johanneswuerbach:
After https://github.com/tildeio/htmlbars/pull/293, this should works now.

* Use ember-cli for testing
* Moved cross-browser testing options into a separate config
* Removed testem dependency


johanneswuerbach:
This is ready now. @rwjblue r?

rwjblue:
@johanneswuerbach - Thanks!

PR #300, opened 2015-02-17
[WIP] Globals + canary bower artifact repo deployment

jayphelps:
I can haz a bower dist? How bout globals?

I'll need someone to help get this past the finish line. @rwjblue put me in the right direction but the global dist fails to load with `Uncaught Error: Could not find module htmlbars-util/handlebars/safe-string` (others inside `htmlbars-util` also will fail to find more `./htmlbars/*`).

Digging through, I expect it has something to do with https://github.com/tildeio/htmlbars/pull/194. Does `amd/htmlbars.amd.js` even work right now as-is? Doesn't for me.

- [ ] Fix `htmlbars-util/handlebars` inline dependency issue
- [ ] Someone created [tildeio/htmlbars-builds](https://github.com/tildeio/htmlbars-builds) or equivalent artifact storage.

AFAIK I can't truly pre-test all the travis -> artifact push stuff, but it's mostly copy-pasta from [ember-qunit](https://github.com/rwjblue/ember-qunit), et al.


Brocfile.js
```diff
@@ -167,16 +167,48 @@ for (var packageName in packages.dependencies) {
...
   var transpiledCjsTests = transpileES6(mergeTrees(testTrees), { type: 'cjs' });
   var movedCjsTests = new Funnel(transpiledCjsTests, {
-    srcDir: packageName+'-tests/',
-    destDir: '/cjs/'+packageName+"-tests/"
+    srcDir: packageName + '-tests/',
+    destDir: '/cjs/' + packageName + '-tests/'
   });
   trees.push(movedCjsTests);
+
+  console.log('>>>' ,packageName);
```
mixonic: Intentional console log?

jayphelps: DOH

mixonic:
@jayphelps the runtime package would possibly used if you wanted to consume HTMLBars as a stand-alone templating engine. In Ember, we provide our own hooks implementation to HTMLBars and thus in many ways the hooks are only used for testing and proof-of-concept. @mmun might be able to say if that would change with the new reactive rendering changes. I know the concept of hooks will change.

I'm unsure if we're stable enough to have a great story here yet, but I would love to see HTMLBars a library used for things beyond Ember.

jayphelps:
@mixonic I *definitely* need `htmlbars-runtime` myself. One of the above tasks mentions I wasn't sure where it would live, globals-wise.

mixonic:
@jayphelps for sure, I'm open to exposing it in an AMD and globals public API. `HTMLBars.Runtime.hooks` seems ok. But as I hope I made clear: I am uncertain of if hooks will even survive some of the changes @wycats and @tomdale are working on.

jayphelps:
@mixonic Yup, got it. We will adapt; Resistance is futile.

lazd:
Definitely interested in seeing this finished :+1:

jayphelps:
@lazd waiting for Glimmer changes to settle. 😀

PR #309, opened 2015-03-16
Removes duplicate code

aymerick:
The `compileSpec` and `template` functions are defined two times in the same file.


rwjblue:
:+1:

PR #310, opened 2015-03-17
Pass options through to AST plugins.

rwjblue:
This allows AST plugins to have more information about the thing being compiled. The intent is for this to be a way to provide compile time template deprecations and still be able to indicate the module path for the template being compiled.

The general idea is that ember-cli-htmlbars will call `precompile` something like:

```javascript
var compiler = require('./bower_components/ember/ember-template-compiler');

var template = compiler.precompile(input, {
  moduleName: 'some-app/path/to-template/stuff'
});
```

And then Ember can have an AST preprocessor that deprecates things (like `{{bind-attr}}`) at compile time, but still provides the template path being compiled. Without the module name tracking down deprecations on compile is extremely difficult.


rwjblue:
Pulled into v0.11.3.

stefanpenner:
Nice

PR #313, opened 2015-03-25
Added destroy method to AttrMorph prototype

taras:
- Added test for presence and type

At suggestion by @stefanpenner


rwjblue:
AttrMorph !== AttrNode

Are you certain that you didn't mean to do this for AttrNode inside Ember
?

stefanpenner:
ya this is the thing we wanted destroy on, but it turns out. Another bug caused it to leak into an area which caused it to be destroyed.

Issue #314, opened 2015-03-26
IE8 <select> Auto selection behavior broken.


rwjblue:
Seems that the auto-select behavior is now broken for IE8.

https://github.com/emberjs/ember.js/blob/beta/packages/ember-views/tests/views/select_test.js#L136-150 is the broken test in Ember.

I believe that this is related to:

https://github.com/tildeio/htmlbars/commit/16436c20e870cf25a454dc9c3e2428057a76adb2
https://github.com/tildeio/htmlbars/commit/58fb9272f12e794634ff083cd10ad408d9cbd157

PR #317, opened 2015-03-27
[WIP] Add <select> auto-selection tests.

rwjblue:
Initial test for https://github.com/tildeio/htmlbars/issues/314.

Hopefully serves as a starting point for @mixonic...


rwjblue:
restarted travis, it was failing for IE8 but that was fixed recently

PR #319, opened 2015-04-02
fix param name

tikotzky:


mmun:
It should be `[ person ]` with array destructuring.

tikotzky:
I fixed to use ES5 style `params[0]` (like the `upcase` above).
Should I rewrite using ES6 style destructuring?

mixonic:
lgtm

PR #320, opened 2015-04-02
Normalize values passed to setPropertyStrict

mixonic:
IE dislikes setting null or undefined values to value and type properties. Several browsers only accept a blank string as an appropriate value for "no url" with src tags.

This brings over logic previously living in Ember.

### DIFF
--- a/packages/dom-helper/lib/main.js
+++ b/packages/dom-helper/lib/main.js
@@ -242,6 +242,14 @@ if (canRemoveSvgViewBoxAttribute){
 }
 
 prototype.setPropertyStrict = function(element, name, value) {
+  if (value === undefined) {
+    value = null;
+  }
+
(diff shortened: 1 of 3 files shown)


packages/dom-helper/lib/main.js
```diff
@@ -242,6 +242,14 @@ if (canRemoveSvgViewBoxAttribute){
 }
 
 prototype.setPropertyStrict = function(element, name, value) {
+  if (value === undefined) {
+    value = null;
+  }
+
+  if (value === null && (name === 'value' || name === 'type' || name === 'src')) {
```
rwjblue: Isn't this dependent on element type?  Should we check and cache to make sure we do the right thing here?

mixonic: Most of this is probably element-type specific, yeah. But I'm also willing to punt on that for now :-p

PR #324, opened 2015-04-07
Fix error for void elements

chancancode:
While looking at #323, I also noticed that the error message was "wrong" (inaccurate), so I fixed it and added a test case for it. Previously, the "void elements cannot have end tags" only work if this happens in the root context.

This is the minimal diff to pass the test. Open to suggestions to make it more elegant.

Input:

```html
<div>
  <input></input>
</div>
```

Before:

```
Closing tag `input` (on line 2) did not match last open tag `div` (on line 1).
```

After:

```
Invalid end tag `input` (on line 2) (void elements cannot have end tags).
```

### DIFF
--- a/packages/htmlbars-compiler/tests/html-compiler-test.js
+++ b/packages/htmlbars-compiler/tests/html-compiler-test.js
@@ -878,11 +878,14 @@ test("A helpful error message is provided for unmatched end tags", function() {
 });
 
 test("A helpful error message is provided for end tags for void elements", function() {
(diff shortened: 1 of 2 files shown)


rwjblue:
LGTM

rwjblue:
@mmun - r?

rwjblue:
@mmun - Can you review when you have a minute?

mmun:
Thanks!

PR #325, opened 2015-04-08
Fix compile and run

chancancode:
Some recent changes broke compile-and-run, I think I fixed it! (see the commit messages for details about each change)

### DIFF
--- a/demos/compile-and-run.html
+++ b/demos/compile-and-run.html
@@ -33,7 +33,6 @@
       var compiler    = requireModule('htmlbars-compiler'),
           DOMHelper   = requireModule('dom-helper').default,
           hooks       = requireModule('htmlbars-runtime').hooks,
-          helpers     = requireModule('htmlbars-runtime').helpers,
           render      = requireModule('htmlbars-runtime').render;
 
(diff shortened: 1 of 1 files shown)


chancancode:
By the way, attempting to invoke a non-existent helper (e.g. putting `{{#if ...}}` in the demo) now blows up with an `undefined.call(...)`. I haven't digged deep enough to know if that is expected or not. I might need to pass something more sophisticated than an empty POJO to `render`, unsure. But before my commit it was blowing up for a different reason (passing `undefined` as `helper`, so it did something like `undefined["if"]` in `lookupHelper`), so this is at least a partial fix.

mmun:
Now that idempotent-rerender has been merged helpers should look the same across all "implementations" of HTMLBars. We could (should?) bake in the basic ones, (`if`, `each`, etc).

chancancode:
Second pass: 

- [*] Fixed test
- [*] `bindSelf` to the data hash instead of copying its keys/values

Should be good to

mmun:
Thanks. We really need to improve the top-level ergonomics :sweat_smile:

PR #327, opened 2015-04-15
Do not write null/undefined properties on initial render.

rwjblue:


packages/morph-attr/lib/main.js
```diff
@@ -4,7 +4,12 @@ import { svgNamespace } from "./dom-helper/build-html-dom";
 import { getAttrNamespace } from "./htmlbars-util";
 
 function updateProperty(value) {
-  this.domHelper.setPropertyStrict(this.element, this.attrName, value);
+  if (this._renderedInitially === true || (value !== undefined && value !== null)) {
```
mixonic: `if (this._renderedInitially === true || !isAttrRemovalValue(value)) {`

rwjblue: Awesome, thank you for pointing out the util function!

rwjblue:
@mixonic r+ in chat

PR #334, opened 2015-05-03
Do not set `.value` property to the current value.

rwjblue:
Allows preserving cursor position, during a re-render when a user is entering text into a bound input.


rwjblue:
r+ from @wycats in chat

rwjblue:
Also, we want to make this less hacky in the future.

PR #335, opened 2015-05-05
Adding params to component hook signature

ef4:
This supports https://github.com/emberjs/ember.js/pull/11028


rwjblue:
Released as 0.13.14.

PR #336, opened 2015-05-05
Add AttrMorph.prototype.clear and AttrMorph.prototype.destroy.

rwjblue:
These methods are called by `renderAndCleanup` (from `template-utils`)
when an AttrMorph is created in a block helpers template:

```javascript
    {{input type="checkbox" checked=checked}}

    {{#if checked}}
      <a href={{"http://google.com"}}>google.com</a>
    {{else}}
      <a href={{"http://yahoo.com"}}>yahoo.com</a>
    {{/if}}
```

In the above example, when `checked` is toggled the `href` AttrMorphs get `.clear()`
and `.destroy()` called on them.

### DIFF
--- a/packages/morph-attr/lib/main.js
+++ b/packages/morph-attr/lib/main.js
@@ -89,6 +89,18 @@ AttrMorph.prototype.getContent = function () {
   return value;
 };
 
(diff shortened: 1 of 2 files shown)


mixonic:
Lgtm

PR #340, opened 2015-05-09
Expose inverse block to component hook

ef4:
This exposes the inverse template to the component hook, and parameterized `bindBlock` and `yield` so they can optionally use other block names (which will presumably just be "inverse" for now).

### DIFF
--- a/packages/htmlbars-runtime/lib/hooks.js
+++ b/packages/htmlbars-runtime/lib/hooks.js
@@ -98,7 +98,7 @@ export function wrap(template) {
 }
 
 export function wrapForHelper(template, env, scope, morph, renderState, visitor) {
-  if (template === null) {
+  if (!template) {
     return {
       yieldIn: yieldInShadowTemplate(null, env, scope, morph, renderState, visitor)
     };
@@ -303,7 +303,7 @@ export function createFreshScope() {
   // because `in` checks have unpredictable performance, keep a
   // separate dictionary to track whether a local was bound.
   // See `bindLocal` for more information.
-  return { self: null, block: null, locals: {}, localPresent: {} };
+  return { self: null, blocks: {}, locals: {}, localPresent: {} };
 }
 
 /**
@@ -426,11 +426,13 @@ export function updateLocal(env, scope, name, value) {
   Corresponds to entering a shadow template that was invoked by a block helper with
   `yieldIn`.
 
-  This hook is invoked with an opaque block that will be passed along to the
-  shadow template, and inserted into the shadow template when `{{yield}}` is used.
+  This hook is invoked with an opaque block that will be passed along
+  to the shadow template, and inserted into the shadow template when
+  `{{yield}}` is used. Optionally provide a non-default block name
+  that can be targeted by `{{yield to=blockName}}`.
 */
-export function bindBlock(env, scope, block) {
-  scope.block = block;
+export function bindBlock(env, scope, block, name='default') {
+  scope.blocks[name] = block;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/htmlbars-runtime/lib/hooks.js
```diff
@@ -909,9 +914,9 @@ export function getCellOrValue(reference) {
   return reference;
 }
 
-export function component(morph, env, scope, tagName, params, attrs, template, visitor) {
+export function component(morph, env, scope, tagName, params, attrs, template, inverse, visitor) {
```
rwjblue: Could we just make this:

```javascript
export function component(morph, env, scope, tagName, params, attrs, templates, visitor) { }
```

Where `templates` is an object that contains `template` and `inverse` properties?  That would in theory align us for future work in Handlebars/HTMLBars for named blocks without changing the component hook signature again...

packages/htmlbars-runtime/lib/hooks.js
```diff
@@ -422,15 +422,18 @@ export function updateLocal(env, scope, name, value) {
...
-  shadow template, and inserted into the shadow template when `{{yield}}` is used.
+  This hook is invoked with an opaque block that will be passed along
+  to the shadow template, and inserted into the shadow template when
+  `{{yield}}` is used. Optionally provide a non-default block name
+  that can be targeted by `{{yield to=blockName}}`.
 */
-export function bindBlock(env, scope, block) {
-  scope.block = block;
+export function bindBlock(env, scope, block, blockName='block') {
+  scope[blockName] = block;
```
wycats: -1 on just stashing this on the scope's root. Can you make a `blocks` member that holds the named blocks? Then it'll be reliably future-proof for other `yield to`

packages/htmlbars-runtime/lib/hooks.js
wycats: :+1:

packages/htmlbars-runtime/lib/hooks.js
ef4: Ok. I was prematurely optimizing.

rwjblue: I think that he means:

```javascript
scope.blocks = scope.blocks || {};
scope.blocks[blockName] = block;
```

So that the blocks (with potentially arbitrary names in the future) don't live directly off of `scope`.

ef4: Yes, I understood, will do.

I originally just cheaped out on an additional object instantiation, since
this is fairly hot code.

packages/htmlbars-runtime/lib/hooks.js
ef4: htmlbars is chock full of functions that accept `template` and `inverse` as separate arguments. To do this right, we would change almost all of them to pass around a `templates` object instead. That looks like a big refactor. I'm going to limit my change to the `component` hook for now.

rwjblue: That makes sense. I will make an issue to track refactoring the rest...

rwjblue: Created: https://github.com/tildeio/htmlbars/issues/341

ef4:
Incorporated feedback from @wycats and @rwjblue, and ready for review.

wycats:
Any chance for tests?

ef4:
If there were any existing tests covering these areas, I would happily have extended them. But there are not, and I don't have sufficient context to create a meaningful test harness from scratch.

rwjblue:
Published as 0.13.17.

PR #344, opened 2015-05-13
manualElement handles SVG properly

salzhrani:
sets namespace if `tagName` is SVG


rwjblue:
Thank you!

mixonic:
Blah, should probably also be detecting the namespace with the contextual element before it does the SVG check. Needed for foreignElement, though few people are going to bump into that.

PR #347, opened 2015-05-19
Do not attempt to redirect null paths

mixonic:
`null` paths were being passed to the `classify` and `handleKeyword` hook. This was obviously not intended, and the fact that code in `classify` and `handleKeyword` is safe for a `null` path was not intentional.

This fixes https://github.com/emberjs/ember.js/pull/11223 in Ember.


rwjblue:
Can you link which bug (so we can reference and close later)?

mixonic:
For sure, updated.

PR #349, opened 2015-05-21
Inline Hook False Value Fix

### LINKED ISSUE #348: inline hook doesn't update morph to false-like values
The below block of code in the inline hook causes issues when a false-like value is returned.

If `morph.lastValue` is `hello` and you want to update the value to a blank value (e.g. null or '') the check `result && result.value` stops the value from actually being updated.

    if (result && result.value) {
      var value = result.value;
      if (morph.lastValue !== value) {
        morph.setContent(value);
      }
      morph.lastValue = value;
    }

In the new Ember {{get}} keyword there is a particular [workaround](https://github.com/emberjs/ember.js/blob/master/packages/ember-htmlbars/lib/keywords/get.js#L28) in place to compensate for this issue.

jmurphyau:
Fixed an issue where the inline hook would not correctly update to a false-like value, instead retaining it's previous value

Fixes #348

### DIFF
--- a/packages/htmlbars-runtime/lib/hooks.js
+++ b/packages/htmlbars-runtime/lib/hooks.js
@@ -655,7 +655,7 @@ export function inline(morph, env, scope, path, params, hash, visitor) {
   var helper = env.hooks.lookupHelper(env, scope, path);
   var result = env.hooks.invokeHelper(morph, env, scope, visitor, params, hash, helper, options.templates, thisFor(options.templates));
 
-  if (result && result.value) {
+  if (result && 'value' in result) {
     var value = result.value;
     if (morph.lastValue !== value) {
       morph.setContent(value);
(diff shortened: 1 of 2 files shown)


jmurphyau:
cc @rwjblue

I got a few JSHint errors which existed in files I wasn't going anywhere near - didn't touch those.. I'm assuming JSHint errors don't stop the CI build from passing like it does in Ember..

rwjblue:
Those JSHint errors are actually from transpiled Handlebars code, but only JSHint failures in files from this repo will fail the build.

mmun:
What's the point of the `value` check at all? Can we remove it?

rwjblue:
@mmun - Helpers can choose to return undefined (the do not actually have to return anything at the moment).

mmun:
@rwjblue I don't understand how that is relevant. In HTMLBars, after the suggested change, that check will always true.

rwjblue:
Replaced by #356.  Thanks again for identifying this @jmurphyau!

PR #353, opened 2015-05-28
detect props that aren't writable, fixes emberjs/ember.js#11221

jayphelps:
Cc/ @stefanpenner 

Working on tests, but want your input on the implementation while I do...

### DIFF
--- a/packages/dom-helper/lib/prop.js
+++ b/packages/dom-helper/lib/prop.js
@@ -9,17 +9,16 @@ export var propertyCaches = {};
 
 export function normalizeProperty(element, attrName) {
   var tagName = element.tagName;
-  var key, cachedAttrName;
   var cache = propertyCaches[tagName];
   if (!cache) {
     // TODO should this be an o_create kind of thing?
     cache = {};
-    for (cachedAttrName in element) {
-      key = cachedAttrName.toLowerCase();
-      if (isSettable(element, cachedAttrName)) {
-        cache[key] = cachedAttrName;
+    for (let key in element) {
+      let lowerKey = key.toLowerCase();
+      if (isSettable(element, key)) {
+        cache[lowerKey] = key;
       } else {
-        cache[key] = UNDEFINED;
+        cache[lowerKey] = UNDEFINED;
       }
     }
     propertyCaches[tagName] = cache;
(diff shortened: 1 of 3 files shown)


packages/dom-helper/lib/prop.js
```diff
@@ -13,11 +16,26 @@ export function normalizeProperty(element, attrName) {
...
   // presumes that the attrName has been lowercased.
-  return cache[attrName];
+  var value = cache[attrName];
+  return value === UNDEFINED ? undefined : value;
+}
+
+function isSettable(element, attrName) {
+  var desc = getOwnPropertyDescriptor(element.constructor.prototype, attrName);
+  if (!desc) { return true; }
+  if (!desc.writable || !desc.hasOwnProperty('value') && typeof desc.set !== 'function') {
```
jayphelps: This is the confusing and important part. We can change the "value" of a property (e.g. what we would get back if we accessed it) if it is *writable OR it doesn't have a `value` AND has a `set` function*. Browser spec guarantees that we will never get back a descriptor that has both a `value` and a `set`. Always mutually exclusive.

stefanpenner:
this may be better then: https://github.com/tildeio/htmlbars/pull/352 but I wonder what examples beyond `form` require this functionality. I suspect in the world of custom WebComponents this may be the case, but i don't fully understand that scenario.

Can you share some further examples to help me understand this?

packages/dom-helper/lib/prop.js
jayphelps: > Property descriptors present in objects come in two main flavors: data descriptors and accessor descriptors. A data descriptor is a property that has a value, which may or may not be writable. An accessor descriptor is a property described by a getter-setter pair of functions. A descriptor must be one of these two flavors; it cannot be both. ~ [Object.defineProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty)

jayphelps: Note, while writing the tests, I just realized the first part `!desc.writable` isn't specific enough because of implicit boolean conversion of `undefined`. it needs to be `desc.writable === false`

jayphelps:
@stefanpenner 

Absolutely. While I wouldn't be surprised if there were more prop/attr pairs in the spec with this behavior, my major concern is indeed custom elements.

```js
class XPerson extends HTMLElement {
  @readonly
  age = 33;
}

document.registerElement('x-person', XPerson);
```

```hbs
<x-person age={{someAge}}></x-person>
```

When I write his out, it actually indeed makes me wonder if properties *truly marked as readonly* should throw an error (or let the browser naturally throw it by writing to it) while properties which simply don't have a setter would instead just write to the attribute.

I can see cases both ways...So I'm unsure.

jayphelps:
Summary: Do we need a distinction between "effectively readonly" and `writeable: false`?

packages/dom-helper/lib/prop.js
```diff
@@ -2,6 +2,9 @@ export function isAttrRemovalValue(value) {
   return value === null || value === undefined;
 }
 
+var getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor;
```
rwjblue: Can you confirm this works on IE8 (which we still have to support)?

https://msdn.microsoft.com/library/dd548686%28v=vs.94%29.aspx

As per that article, getOwnPropertyDescroptor does work in IE8 for "DOM Objects", so it may be perfectly fine, I just want to confirm.

rwjblue: Thinking more about this: we have IE8 in CI so if tests are green (and we are testing this path), we are good.

stefanpenner:
the `input[type]` scenario doesn't appear to be handled by this.

jayphelps:
@stefanpenner If I'm following that thread correctly, IMO that's an unrelated case of "prop doesn't *always* reflect to attribute". The code for that might be in the same place, but unrelated concepts.

jayphelps:
Okay, I think I understand better now, the issue is actually the fact that is errors, not that it doesn't reflect. (still unrelated, but yeah, fix is in the same code path. I'm attempting a PR right now)

stefanpenner:
> Okay, I think I understand better now, the issue is actually the fact that is errors, not that it doesn't reflect. (still unrelated, but yeah, fix is in the same code path. I'm attempting a PR right now)
some funkyness for sure.
:+1:

rwjblue:
Can haz rebase?

packages/dom-helper/lib/prop.js
```diff
@@ -15,18 +15,17 @@ export function normalizeProperty(element, attrName) {
...
-        cache[key] = UNDEFINED;
+        cache[lowerKey] = UNDEFINED;
       }
     }
     propertyCaches[tagName] = cache;
   }
 
-  // presumes that the attrName has been lowercased.
-  var value = cache[attrName];
+  var value = cache[attrName.toLowerCase()];
```
jayphelps: I need @mixonic to chime in why before this "presumed" attrName had been lowercased, when in fact it never was (at least in my testing). From what I could tell, the cache always missed when the `attrName !== attrName.toLowerCase()`. Was that for some reason intended?

packages/dom-helper/lib/prop.js
```diff
@@ -59,5 +58,49 @@ function isSettable(element, attrName) {
     }
   }
 
+  // Properties can be effectively read-only two ways.
+  // If actually marked as writable = false, an exception is thrown if you attempt
+  // to assign. If it's simply missing a setter, it silently just doesn't
+  // assign anything. Both cases we will defer to setAttribute instead
+  var desc = getPropertyDescriptor(element, attrName);
+  if (!desc) { return true; }
+  if (desc.writable === false || !desc.hasOwnProperty('value') && typeof desc.set !== 'function') {
```
jayphelps: @rwjblue and I are discussing this in Slack...but this `typeof desc.set !== 'function'` fails in Safari due to this outstanding bug: https://bugs.webkit.org/show_bug.cgi?id=49739. tl;dr Safari is the only browser that doesn't return full descriptors for host element properties.

Chrome:
image

Safari:
image

jayphelps:
@rwjblue @stefanpenner @mixonic Given the discovery here https://github.com/tildeio/htmlbars/pull/353#discussion_r32170415, I've updated the solution to exclusively use a blacklist approach for *all native elements*. For Custom Elements, the robust solution is used for "ideal" interop. Whether or not setting the attribute when the property is not settable for Custom Elements is the right thing to do (vs. erroring or just not doing anything) is a bigger question of how third party web component libs like polymer, graffiti, etc expect things.

Please review and provide thoughts.

jayphelps:
To further complicate things...in some ways we're going down the same path that React does. Check this out: https://github.com/facebook/react/blob/4c3e9650ba6c9ea90956a08542d9fa9b5d72ee88/src/renderers/dom/shared/HTMLDOMPropertyConfig.js

stefanpenner:
Find me at the meetup tonight. I believe we have a much better long term fix. Although this may still be a nice interim solution.

Robdel12:
Any updates on this? :)

jayphelps:
@Robdel12 My latest understanding is that the fundamental way we choose whether we use the `el.property = value` vs `setAttribute` is likely to change. That said, @stefanpenner and I believed the issue would still exist for people choosing the wrong syntax (which doesn't exist yet) and we'd try and still fix it with this PR, but throw a warning of some kind.

I'm not the authority on this, just my understanding of it.

The idea is that `<x-foo attr="{{bar}">` would use `setAttribute because you put the mustache in quotes (HTML attributes are, after-all *always* strings) but if you omit the quotes `<x-foo prop={{bar}>` it would then use the same "props first if exists, setAttribute if not" pattern currently in place.

I'm personally in the camp of "this is going to confuse the hell out of people" because: previously we didn't really make it clear to devs that we were doing a props first, setAttribute second to begin with, most devs I've queried understandably don't realize there's a different between properties and attributes (I'm not joking), and *maybe* it won't be clear when to quote your 'staches and when not to. I'm super attached to this subject though, so I wouldn't be surprised if I was wrong about one or all of my concerns. I think it needs community input, likely an RFC.

### tl;dr
I hope this PR gets merged so people can move on, at least for now.

Robdel12:
Thanks for the update! That helped me understand what's going on here.

jamesarosen:
I don't know if it helps, but here's a [JSBin](http://jsbin.com/rolobi/edit?html,js,output) that shows that

 1. setting `form` via `prop` doesn't take
 2. setting `form` via `attr` will cause `<select>.form` to be the new `<form>`
 3. even if you set via `attr`, submitting the form indicated by `form='...'` doesn't incorporate the field, which is the point of that attribute

jayphelps:
Superseded by #374

PR #354, opened 2015-05-28
attempt at fixing SVG integration elements

salzhrani:
This adds an extra `dom.setNamespace` which fixes the integration object issues but fails a test that checks for unnecessary `dom.setNamepsace` calls. not the best approach but it works.


mixonic:
@salzhrani can't this just be part of the manual element? Set the namespace to null after creating the element if the element is an SVG?

salzhrani:
@mixonic the problem is that the template is rendered without information about the `tagName` of the component, thus the namespace is not set to the svg namespace, consequently  not triggering a namespace change because of the integration element. Even if you reset the namespace in `manualElement` it gets set again to the svg namespace  [here](https://github.com/tildeio/htmlbars/blob/master/packages/htmlbars-runtime/lib/render.js#L23)

mixonic:
> the problem is that the template is rendered without information about the tagName of the component

Should never be true. When a template is rendered, it *must* be provided a contextual element. If we are not passing the contextual element it is absolutely a regression.

Maybe this regressed in the Glimmer work.

PR #356, opened 2015-05-30
Allow falsey values from helpers.

rwjblue:
Thanks to @jmurphyau for the fix (in https://github.com/tildeio/htmlbars/pull/349)!


jmurphyau:
👍 awesome

Can you /cc me when Ember is updated with this version - I'll update the get helper to remove the workaround

PR #358, opened 2015-05-31
Helper streams require get

mixonic:
If `invokeHelper` is to return a stream, we must ensure that stream always has a value generated. Sometimes a stream's compute function may yield a block instead of generate a value.

This forces compute to happen once.

There is probably a better way to achieve the base goal here. For example, I should likely only link nodes that a new-style `Ember.Helper`, where they cannot interact with `yield` etc.

See also: https://github.com/mixonic/ember.js/pull/8


mmun:
LGTM

mixonic:
Cleaned this up to only call the value hook when the value is used. Ensuring that a helper using internals (yielding blocks etc) will be run is the responsibility of `invokeHelper` in Ember.

PR #363, opened 2015-06-07
Force <input> `list` to be set as an attribute.

rwjblue:
Addresses:

* https://github.com/emberjs/ember.js/issues/10908
* https://github.com/emberjs/ember.js/issues/11364


mixonic:
Gracias.

PR #370, opened 2015-06-17
Remove IE8 workarounds

mmun:


rwjblue:
LGTM 

@mixonic - r?

mmun:
@mixonic Do you mind taking a peeking and checking if I missed anything? I only removed code near comments that explicitly mentioned IE8.

mixonic:
:clap: this is great. Thanks @mmun!

stefanpenner:
:shipit:

PR #378, opened 2015-07-06
Implement new visitor

mmun:
Implements a new visitor that fully traverses the AST. It was designed in the spirit of Babel's transform. Usage:

```js
import { traverse } from "htmlbars-syntax";

let ast = parse(`{{name-tag name=(upcase user.name)}}`);

traverse(ast, {
  PathExpression(node) {
    console.log(node.original); // Logs "name-tag", "upcase", "user.name"
  }
});
```

Alternatively, you can use the enter/exit format to capture when the node is entered and exited:

```js
import { traverse } from "htmlbars-syntax";

let ast = parse(`...`);
let depth = -1;

traverse(ast, {
  Program: {
    enter() { depth++; },
    exit() { depth--; }
  }
});
```

### DIFF
--- a/packages/htmlbars-syntax/lib/main.js
+++ b/packages/htmlbars-syntax/lib/main.js
@@ -1,9 +1,11 @@
 import Walker from "./htmlbars-syntax/walker";
 import builders from "./htmlbars-syntax/builders";
 import parse from "./htmlbars-syntax/parser";
+import traverse from "./htmlbars-syntax/traversal/traverse";
 
 export {
   Walker,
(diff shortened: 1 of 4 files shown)


mmun:
There are more features coming for this visitor:

  - aliases (e.g. the `Statement` key is visited for any kind of Statement)
  - replacement API (calling `this.remove()` or `this.replace(nodes)` inside the node handlers)

mixonic:
Nice!!! :star: :star: :star:

PR #379, opened 2015-07-06
Cleanup syntax module a bit

mmun:
- Use the root `parse` export on `htmlbars-syntax` instead of `preprocess`
- Prefix QUnit modules to make filtering easier


packages/htmlbars-syntax/tests/loc-node-test.js
```diff
@@ -1,6 +1,6 @@
 import { parse } from "../htmlbars-syntax";
 
-QUnit.module("Parser - Location Info");
+QUnit.module("[htmlbars-syntax] Parser - Location Info");
```
stefanpenner: :+1:

PR #382, opened 2015-07-08
Implement replacement API

mmun:
You can now return `null`, a node or an array of nodes inside of the `enter` and `exit` events to replace the current node. Returning `null` is equivalent to returning an empty array. You can only replace a node with multiple nodes (or zero nodes) if it is part of an array.

If you return a replacement in the `enter` event then replacement's child nodes **will not** be visited (the user created the nodes, so there's no point) and moreover the `exit` event for the current node **will not** be called (this would just be confusing, since the node has been replaced possibly by multiple nodes).

Here is an illustrative example from the tests:

```js
QUnit.test('Inside of a block', () => {
  let ast = parse(`{{y}}{{#w}}{{x}}{{y}}{{z}}{{/w}}`);

  traverse(ast, {
    MustacheStatement(node) {
      if (node.path.parts[0] === 'y') {
        return [
          b.mustache('a'),
          b.mustache('b'),
          b.mustache('c')
        ];
      }
    }
  });

  astEqual(ast, `{{a}}{{b}}{{c}}{{#w}}{{x}}{{a}}{{b}}{{c}}{{z}}{{/w}}`);
});
```

### DIFF
--- a/packages/htmlbars-syntax/lib/builders.js
+++ b/packages/htmlbars-syntax/lib/builders.js
@@ -3,7 +3,7 @@
 export function buildMustache(path, params, hash, raw, loc) {
   return {
     type: "MustacheStatement",
-    path: path,
+    path: buildPath(path),
     params: params || [],
     hash: hash || buildHash([]),
     escaped: !raw,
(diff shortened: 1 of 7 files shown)


stefanpenner:
cc @abuiles

stefanpenner:
rad!

zackthehuman:
This is great! This addresses my needs perfectly. Thanks for putting in the time to do this!

PR #390, opened 2015-07-17
[Bugfix] Adjusting issue #11727 from Ember.js that is relation related with htmlbars.

rodrigo-morais:
Hello,

I'm correcting the bug #11727 from Ember.js that is related with HTMLBars. The bug happens when a component is created using angle brackets and a property receive a value as a boolean. The HTMLBars is not treating this value as a boolean.

### DIFF
--- a/packages/htmlbars-syntax/lib/builders.js
+++ b/packages/htmlbars-syntax/lib/builders.js
@@ -96,6 +96,14 @@ export function buildText(chars, loc) {
   };
 }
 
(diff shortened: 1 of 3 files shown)


mixonic:
@rodrigo-morais this would bend some rules about trying to match HTML semantics pretty hard. I believe we suggest users do `<a-component foo={{true}}>` or `<a-component foo={{false}}>` in these cases.

Thank you for taking the time to open this PR, and please let me know if I've missed anything.

PR #393, opened 2015-07-22
handle duplicate keys

krisselden:
needs test, verified in Ember


ef4:
See failing test in https://github.com/emberjs/ember.js/pull/11949

rwjblue:
emberjs/ember.js#11949 will be fixed by tildeio/htmlbars#398

PR #397, opened 2015-08-01
Implement AST printer

code0100fun:
This is the most basic AST printer. It does **not** reproduce the exact source that generated the AST.

That is to say, these tests don't always pass:
`equal(print(parse(template)), template)`
`deepEqual(parse(print(ast)), ast)`

Currently, all statements and expressions are handled with the exception of comments, so the new source will have the same semantic meaning. However, certain formatting is lost.

Formatting:
- [x] newlines*
- [x] tabs
- [x] arbitrary indent sizes
- [x] blank lines
- [ ] newlines after opening or closing a block expression (seems to be a Handlebars issue)
- [ ] newlines between element attributes
- [ ] newlines between mustache attributes
- [ ] arbitrary spacing between element tag, attributes, and angle brackets
- [ ] arbitrary spacing between mustache path, params, hash, and attributes

Statements:
- [ ] Handlebars comment (stripped out of AST by Handlebars)

### DIFF
--- a/packages/htmlbars-syntax/tests/generation/print-test.js
+++ b/packages/htmlbars-syntax/tests/generation/print-test.js
@@ -1,0 +1,101 @@
+import { parse, print, builders } from '../../htmlbars-syntax';
+
+const b = builders;
+
+function printEqual(template) {
+  const ast = parse(template);
+  equal(print(ast), template);
+}
+
+QUnit.module('[htmlbars-syntax] Generation - printing');
+
+test('ElementNode: tag', function() {
+  printEqual('<h1></h1>');
+});
+
+test('ElementNode: nested tags with indent', function() {
+  printEqual('<div>\n  <p>Test</p>\n</div>');
+});
+
+test('ElementNode: attributes', function() {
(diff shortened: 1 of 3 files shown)


packages/htmlbars-syntax/tests/generation/print-test.js
```diff
@@ -0,0 +1,101 @@
...
+test('BlockStatement: multiline', function() {
+  printEqual('<ul>{{#each foos as |foo|}}\n  {{foo}}\n{{/each}}</ul>');
+});
+
+test('BlockStatement: inline', function() {
+  printEqual('{{#if foo}}<p>{{foo}}</p>{{/if}}');
+});
+
+test('UndefinedLiteral', function() {
+  const ast = b.program([b.mustache(b['undefined']())]);
```
mmun: b.undefined is fine.

mmun:
This should match the style if the codebase. Use `let`, not `const`. `if (` not `if(`. The massive switch statement should be rewritten as an object lookup (similar to the visitor, parser, compiler, etc.).

PR #407, opened 2015-09-02
Properly visit MorphLists in visitChildren.

rwjblue:
`visitChildren` is used internally to visit all `Morph`'s and call a callback on each one found.  The most common usage of `visitChildren` in HTMLBars itself is for clearing morphs before rendering a new template.

Prior to this change, if we encounter a node with `morphList` we:

* add the `morphList` to the list of nodes to traverse
* call the provided callback
* traverse the `MorphList` and add all morphs found there to the list of
  morphs to call the callback on

In general that is fine, except when the callback itself (in step 2 above) is mutating the `morph` and preventing the `MorphList` from being traversed. This is the case when changing an `{{if}}` with a `{{each}}` from the template to the inverse block, in that case the callback is calling `MorphList#clear()` which nullifies the `MorphList` but does not call the various hooks (`didDestroyNode` specifically) so any morphs would be leaked.

This change tweaks `visitChildren` to unroll a `MorphList` upon encountering it (instead of waiting util after the callback has been called on it). The included test demonstrates the memory leak that was occurring in Ember and that it is properly fixed.

### DIFF
--- a/packages/htmlbars-compiler/tests/dirtying-test.js
+++ b/packages/htmlbars-compiler/tests/dirtying-test.js
@@ -462,6 +462,34 @@ test("Pruned render nodes invoke a cleanup hook when replaced", function() {
   strictEqual(destroyedRenderNode.lastValue, 'Nothing', "The correct render node is passed in");
 });
 
+test("MorphLists in childNodes are properly cleared", function() {
+  var object = {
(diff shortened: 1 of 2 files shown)


rwjblue:
r+ from @mmun in chat

PR #400, opened 2015-08-03
[BUGFIX] add `autocorrect` attr override because Mobile Safari hates us

jayphelps:
fixes emberjs/ember.js#11869, Cc/ @stefanpenner

PR #419, opened 2015-09-29
Bump version

chadhietala:
Would like to get a downstream canary where #400 is fixed.


rwjblue:
released manually (after checking with @chadhietala)

Closing...

Issue #420, opened 2015-09-30
Syntax Error in Android Browser API Level <= 15


davidgoli:
Admittedly this is an older browser, affecting as of this writing 8% of the Android market, but I thought I'd submit this for the record:

- open an HTMLBars-dependent client app (eg. an Ember CLI >= 1.13 app) in an Android browser running API level 15 or below (aka Ice Cream Sandwich, v4.0.4) - including an emulator

result: JavaScript syntax error: "Unexpected strict mode reserved word"

Tracked the offending line down to: https://github.com/tildeio/htmlbars/blob/0d883d7dba4e4e7dc476dd7b46a45781a05b6665/packages/htmlbars-runtime/lib/hooks.js:107

Related issue: https://github.com/cujojs/when/issues/89

Does not repro on Android browsers >= API level 16.

rwjblue:
Would you mind sending in a PR quoting 'yield'?

davidgoli:
Done. Hopefully uglifiers won't strip the quotes...

mmun:
Shouldn't babel be handling this?

davidgoli:
I can confirm that this change does indeed fix this issue.

davidgoli:
But it does seem like a nice-to-have for the Babel processor.

PR #421, opened 2015-09-30
quote 'yield' in lib/hooks

davidgoli:
Addresses https://github.com/tildeio/htmlbars/issues/420


rwjblue:
Can you add line comments mentioning yield as a reserved word on old android? I'm afraid that otherwise I might forget and remove this (since it looks "odd")...

davidgoli:
@rwjblue done

davidgoli:
wish there was a reliable way to test this...

rwjblue:
We do test against sauce labs, not sure if they support these android versions though...

rwjblue:
Sorry for the delay here. I guess we should release an updated version for Ember (for likely inclusion in Ember 2.2.1)

Issue #305, opened 2015-03-01
Question: Can we get rid of the empty spaces in the compiled HTML?


martinthogersen:
A normal template, which looks like this:

```hbs
<table class="table table-striped table-hover">
  <thead>
    <tr>
      {{#each columns as |column|}}
        <th>
          {{column.label}}
        </th>
      {{/each}}
    </tr>
  </thead>
  <tbody>
    <tr>
      {{#each itemColumns as |column|}}
        <td>
          {{format-value column.value}}
        </td>
      {{/each}}
    </tr>
  </tbody>
</table>
```

Is compiled to this:

```js
var el0 = dom.createDocumentFragment();
var el1 = dom.createElement("table");
dom.setAttribute(el1,"class","table table-striped table-hover");
var el2 = dom.createTextNode("\n  ");
dom.appendChild(el1, el2);
var el2 = dom.createElement("thead");
var el3 = dom.createTextNode("\n    ");
dom.appendChild(el2, el3);
var el3 = dom.createElement("tr");
var el4 = dom.createTextNode("\n");
dom.appendChild(el3, el4);
var el4 = dom.createComment("");
dom.appendChild(el3, el4);
var el4 = dom.createTextNode("    ");
dom.appendChild(el3, el4);
dom.appendChild(el2, el3);
var el3 = dom.createTextNode("\n  ");
dom.appendChild(el2, el3);
dom.appendChild(el1, el2);
var el2 = dom.createTextNode("\n  ");
dom.appendChild(el1, el2);
var el2 = dom.createElement("tbody");
var el3 = dom.createTextNode("\n    ");
dom.appendChild(el2, el3);
var el3 = dom.createElement("tr");
var el4 = dom.createTextNode("\n");
dom.appendChild(el3, el4);
var el4 = dom.createComment("");
dom.appendChild(el3, el4);
var el4 = dom.createTextNode("    ");
dom.appendChild(el3, el4);
dom.appendChild(el2, el3);
var el3 = dom.createTextNode("\n  ");
dom.appendChild(el2, el3);
dom.appendChild(el1, el2);
var el2 = dom.createTextNode("\n");
dom.appendChild(el1, el2);
dom.appendChild(el0, el1);
return el0;
```

Would it be possible to remove all the "empty" `createTextNode` statements? That would save a bit of space in big projects, and make the HTML content shorter, when Ember fastboot is ready to go.

mmun:
[...]

PR #424, opened 2015-10-02
test to show that stripping whitespace after comments is possible

raycohen:
Adding `{{!~}}` to the end of a line in a htmlbars template file allows suppressing whitespace between that comment and the first non-whitespace character on the next line.

In my case this is useful when converting templates created in emblem (which by default suppresses whitespace) to htmlbars, without needing to manually squish all the dom tags together.

This behavior wasn't documented or tested anywhere.  Hence this PR.

related: https://github.com/tildeio/htmlbars/issues/305


mmun:
Thanks.

PR #359, opened 2015-06-03
update ember-cli

stefanpenner:

PR #425, opened 2015-10-06
Add temporary workaround for SimpleDOM

chancancode:
** Requires https://github.com/krisselden/morph-range/pull/8 **

The current parseHTML implementation requires some DOM APIs not implemented in SimpleDOM. This commit adds a temporary workaround to support using SimpleDOM on the server.

In the future, we will move towards using `insertAdjacentHTML` and `innerHTML=` as the main APIs and remove the need of `parseHTML` altogether.

This commit also reverts a previous commit that introduced the `setMorphHTML` method, which does exactly what `Morph#setHTML` does.

(I added a test, but it's not currently running because we disabled all the node tests since https://github.com/tildeio/htmlbars/pull/359... @stefanpenner ? Anyhow, this particular change is very easy to verify by :eyes:, so it's not critical to re-enable them just to merge this PR.)


chancancode:
@krisselden @wycats can you take a look? :eyes:

Kuzirashi:
Hey, what's the status of this? :) @chancancode

PR #428, opened 2015-10-07
Raise an error for script tags

### LINKED ISSUE #427: Using a script tag should throw an error
If anyone's up for it, here's what to do!

1. Write a function `validateStartTag` and call it in `finishStartTag`: https://github.com/tildeio/htmlbars/blob/master/packages/htmlbars-syntax/lib/parser/tokenizer-event-handlers.js#L80-L86. It should look similar to the existing validation in `finishEndTag`: https://github.com/tildeio/htmlbars/blob/master/packages/htmlbars-syntax/lib/parser/tokenizer-event-handlers.js#L95.
2. `validateStartTag` should throw if the tag name is `script` with a helpful error message (Explanation, line number, etc.)
3. Write a test in https://github.com/tildeio/htmlbars/blob/master/packages/htmlbars-syntax/tests/parser-node-test.js.

g13013:
fixes #427

### DIFF
--- a/packages/htmlbars-syntax/lib/parser/tokenizer-event-handlers.js
+++ b/packages/htmlbars-syntax/lib/parser/tokenizer-event-handlers.js
@@ -203,32 +203,24 @@ function assembleConcatenatedValue(parts) {
 }
 
 function validateStartTag(tag, tokenizer) {
-  var error;
-
-  // no support for <script> tags
+  // No support for <script> tags
   if (tag.name === "script") {
-    error = "SCRIPT tags are not allowed in HTMLBars templates (line " + tokenizer.tagLine + ")";
+    throw new Error("`SCRIPT` tags are not allowed in HTMLBars templates (on line " + tokenizer.tagLine + ")");
   }
-
-  if (error) { throw new Error(error); }
 }
 
 function validateEndTag(tag, element, selfClosing) {
-  var error;
-
   if (voidMap[tag.name] && !selfClosing) {
     // EngTag is also called by StartTag for void and self-closing tags (i.e.
     // <input> or <br />, so we need to check for that here. Otherwise, we would
     // throw an error for those cases.
-    error = "Invalid end tag " + formatEndTagInfo(tag) + " (void elements cannot have end tags).";
+    throw new Error("Invalid end tag " + formatEndTagInfo(tag) + " (void elements cannot have end tags).");
   } else if (element.tag === undefined) {
-    error = "Closing tag " + formatEndTagInfo(tag) + " without an open tag.";
+    throw new Error("Closing tag " + formatEndTagInfo(tag) + " without an open tag.");
   } else if (element.tag !== tag.name) {
-    error = "Closing tag " + formatEndTagInfo(tag) + " did not match last open tag `" + element.tag + "` (on line " +
-            element.loc.start.line + ").";
+    throw new Error("Closing tag " + formatEndTagInfo(tag) + " did not match last open tag `" + element.tag + "` (on line " +
+            element.loc.start.line + ").");
   }
-
-  if (error) { throw new Error(error); }
 }
 
 function formatEndTagInfo(tag) {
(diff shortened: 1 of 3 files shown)


packages/htmlbars-syntax/lib/parser/tokenizer-event-handlers.js
```diff
@@ -200,6 +202,17 @@ function assembleConcatenatedValue(parts) {
   return b.concat(parts);
 }
 
+function validateStartTag(tag, tokenizer) {
+  var error;
+
+  // no support for <script> tags
+  if (tag.name === "script") {
+    error = "SCRIPT tags are not allowed in HTMLBars templates (line " + tokenizer.tagLine + ")";
```
stefanpenner: Why not just throw in this branch

mmun: It's matching the style of the other validation function.

g13013: since it's a validation function, I thought other validation might be added in the future! rewrite it ?

mmun:
Does this catch <SCRIPT> with capitals? We may want to use change the comparison to use `tag.toLowerCase().name`.

packages/htmlbars-syntax/lib/parser/tokenizer-event-handlers.js
mmun: I agree with @stefanpenner, we should change both the validation functions to simply throw immediately.

g13013:
@mmun I fixed  the test to only validate the line number and rebased in the first commit. I also cleaned both functions in the second commit.

I am unsure about uppercase script tag comparison, it is used many times throughout the code and I think, if there is a problem with it, it needs an another PR to fix it everywhere!

timkendrick:
Just seen this PR - is there any chance this behavior can be made optional for specific use cases?

I'm using HTMLBars to render templates on the server-side using simple-dom, so there's no danger of script injection etc. It's working great with the current npm version (0.14.6), and it'd be a real shame for the next update to break all my templates.

Ultimately I'd love it if the `<script>` content could be parsed as a normal text node until it reaches the `</script>` tag, but failing this is there a way to at least prevent throwing this error and retain the current behavior?

I'm fine with monkey-patching the parser at runtime if that's relatively straightforward, and can control the script content to make sure it doesn't confuse the parser, but I worry that I'd have to abandon HTMLBars altogether if it outright vetoes any templates containing `<script>` tags... :/

mixonic:
@timkendrick I think you agree the re-rendering semantics are pretty wacky for something like this. I doubt we should be diverging server-side and client-side capabilities of the templating engine.

I don't suggest you monkey-patch the parser, but you could easily process the simple-dom after rendering (or use the `didRender` hook in Ember) to attach or execute code as you see fit. That should not be very challenging to implement.

> I worry that I'd have to abandon HTMLBars altogether if it outright vetoes any templates containing \<script\> tags... :/

Right, you should be able to do it against the simple-dom directly. I doubt that wrinkle is a good reason to dump all the other features provided by HTMLBars :wink: but I hear the frustration.

timkendrick:
Thanks for the quick response @mixonic.

I'd assume the re-rendering semantics would be relatively straightforward – i.e. if the content changes, update the contents of the `<script>` tag – if it doesn't change, leave it as is. I would have thought this would tie in well with the general idempotence theme of HTMLBars.

If it were me, I'd leave the re-rendering semantics up to the browser: I've just done a few tests, and it looks like the browser won't reexecute the script if you change its `innerHTML` content once it's been attached to the document. This would be unfortunate if you're planning on building something like JSBin that dynamically reevaluates the code at runtime, but that's a very specific use case that I'm sure that could be worked around with a `{{script}}` helper etc.

You're right though – I was being a bit melodramatic in my last paragraph, I'm sure there are ways around this and I'd hate to get rid of HTMLBars. I'm not using Ember, so can you advise on what the best way would be to ensure these script tags get inserted into the correct part of the DOM? Should I make a `{{script}}` helper that acts as a marker node, and then post-process the simple-dom output? It'd be awesome if I could define some kind of `<script>` component that provides the expected behavior without having to modify the template's source HTML, but I'm guessing that this wouldn't make it past the tokenization stage due to the ban on `<script>` elements? I'd be really grateful for some pointers here!

timkendrick:
Also completely agree that you don't want separate server-side and client-side behavior – I reckon in either case, you'd want HTMLBars to just slap the script content in the `innerHTML` and leave it up to the environment to deal with it as it sees fit!

PR #429, opened 2015-10-07
[PERF] Don't Attempt To Sanitive Unsanitiveable

chadhietala:
This adds an identity check for falsy values before the value is
sanitized. Currently we do work to sanitize things like `undefined`.

Before:
screen shot 2015-10-07 at 11 56 21 am

After:
screen shot 2015-10-07 at 11 56 45 am


stefanpenner:
such `Unsanitiveable`

packages/morph-attr/lib/main.js
```diff
@@ -110,7 +110,7 @@ AttrMorph.prototype.setContent = function (value) {
   if (this.lastValue === value) { return; }
   this.lastValue = value;
 
-  if (this.escaped) {
+  if (this.escaped && this.valueIdentity(value)) {
```
stefanpenner: doesn't this just just mean all truthy things are left unescaped ?

stefanpenner:
needs señor tests.

packages/morph-attr/lib/main.js
chadhietala: Maybe I don't understand what is going on here but if should escape e.g. `this.escape === true`, then the value we are trying to escape should be truthy. Currently we run the block with `undefined` or strings. I added the function instead of doing

```
if (this.escaped && value) {

}
```

stefanpenner: ah, this aims to skips escaping if the value is falsy. As we believe no falsey values will need escaping.

stefanpenner: Whats the reason for the method?

chadhietala:
Not really sure how to test this as the sanitization function is local. Not sure how helpful the test is.

packages/morph-attr/tests/attr-morph-test.js
```diff
@@ -17,6 +17,20 @@ test("can update a dom node", function(){
   equal(element.getAttribute('id'), 'twang', 'id attribute is set');
 });
 
+test("setting content to undefined calls _update with undefiend", function(){
```
seanjohnson08: Typo: "undefiend" should be "undefined"

PR #430, opened 2015-10-08
Update some build modules and fix jshint and bower errors 

g13013:
- Updated:
  - ember-cli
  - emberjs-build
  - broccoli-jshint
  - broccoli-funnel

- migrated `Brocfile.js` to `ember-cli-build.js`
- made JsHint ignore `morph-range` package
- fixed bower's `extra-resolution Unnecessary resolution: qunit#~1.18.0` warning


rwjblue:
@chancancode mentioned previously that updating ember-cli caused issues so that change was reverted.  I'd like for him to confirm that those issues are fixed here (I don't believe he elaborated on what they were in the revert PR).

g13013:
If it's the case and if you agree we could keep the build errors then

g13013:
Even though I am asking myself if ember-cli is really necessary to build htmlbars?

g13013:
I close for now, as things are not yet clear

PR #437, opened 2015-12-03
Adds Node-compatible protocolForURL

tomdale:
The current implementation of the DOM helper’s `protocolForURL` relies
on the implicit parsing and normalization that the browser’s DOM does
when setting the A element’s `href` property.

However, the DOM helper is often used with subset DOM implementations
like simple-dom. In those cases, we cannot rely on this normalization.

This commit detects when the DOM Helper is running in Node and uses
the built-in URL package to parse the protocol instead.

/cc @wycats @chancancode @rwjblue @stefanpenner

### DIFF
--- a/packages/dom-helper/lib/main.js
+++ b/packages/dom-helper/lib/main.js
@@ -171,7 +171,7 @@ function DOMHelper(_document){
   this.canClone = canClone;
   this.namespace = null;
 
-  installProtocolForURL(this);
+  installEnvironmentSpecificMethods(this);
 }
 
 var prototype = DOMHelper.prototype;
@@ -571,7 +571,7 @@ prototype.parseHTML = function(html, contextualElement) {
 var URL;
 var parsingNode;
 
-function installProtocolForURL(domHelper) {
+function installEnvironmentSpecificMethods(domHelper) {
   var protocol = browserProtocolForURL.call(domHelper, 'foobar:baz');
 
   // Test to see if our DOM implementation parses
@@ -580,13 +580,26 @@ function installProtocolForURL(domHelper) {
     // Swap in the method that doesn't do this test now that
     // we know it works.
     domHelper.protocolForURL = browserProtocolForURL;
-  } else {
+  } else if (typeof module === 'object' && typeof module.require === 'function') {
     // Otherwise, we need to fall back to our own URL parsing.
     // Global `require` is shadowed by Ember's loader so we have to use the fully
     // qualified `module.require`.
     URL = module.require('url');
     domHelper.protocolForURL = nodeProtocolForURL;
+  } else {
+    throw new Error("DOM Helper could not find valid URL parsing mechanism");
   }
+
+  // A SimpleDOM-specific extension that allows us to place HTML directly
+  // into the DOM tree, for when the output target is always serialized HTML.
+  if (domHelper.document.createRawHTMLSection) {
+    domHelper.setMorphHTML = nodeSetMorphHTML;
+  }
+}
+
+function nodeSetMorphHTML(morph, html) {
+  var section = this.document.createRawHTMLSection(html);
+  morph.setNode(section);
 }
 
 function browserProtocolForURL(url) {
--- a/packages/dom-helper/lib/main.js
+++ b/packages/dom-helper/lib/main.js
@@ -1,3 +1,5 @@
+/*globals module*/
+
 import Morph from "./htmlbars-runtime/morph";
 import AttrMorph from "./morph-attr";
 import {
@@ -168,6 +170,8 @@ function DOMHelper(_document){
   }
   this.canClone = canClone;
   this.namespace = null;
+
+  installProtocolForURL(this);
 }
 
 var prototype = DOMHelper.prototype;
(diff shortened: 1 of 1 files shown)


tomdale:
It looks like CJS commits were disabled in https://github.com/tildeio/htmlbars/commit/0c51a428dd0aa8358c075d73d82fd90df9d6536f but no rationale for why was provided in that commit. Presumably it was for a good reason, but it means I couldn't write a test for this case.

I think it's OK because this is implicitly tested in Ember anyway and this implementation of the DOM Helper is EOL once Glimmer2 lands.

packages/dom-helper/lib/main.js
```diff
@@ -567,13 +569,26 @@ prototype.parseHTML = function(html, contextualElement) {
 var parsingNode;
 
 // Used to determine whether a URL needs to be sanitized.
-prototype.protocolForURL = function(url) {
-  if (!parsingNode) {
-    parsingNode = this.document.createElement('a');
-  }
+// In Node.js environments, we can't rely on the DOM correctly parsing URLs.
+// Instead, we'll fall back to Node's built-in URL parsing.
+if (typeof process === 'object' && process + '' === '[object process]') {
```
stefanpenner: does this work in node-webkit?

stefanpenner:
> It looks like CJS commits were disabled in 0c51a42 but no rationale for why was provided in that commit. Presumably it was for a good reason, but it means I couldn't write a test for this case.

They where breaking the build, and I did not have time to fix them.

packages/dom-helper/lib/main.js
tomdale: Unknown, but looks like this might be a better test: https://github.com/jnath/nwjs-video-player/blob/159c0d487159846664782979d46bbd61857fb429/src/app/js/nw.js#L1

chancancode:
seems good! :+1:

packages/dom-helper/lib/main.js
chancancode: cc @wycats :eyes:

tomdale:
@stefanpenner 
> Maybe someone else can

Basically EOL at this point, not sure it's worth anyone putting effort in here that could be put into glimmer.

stefanpenner:
> Basically EOL at this point, not sure it's worth anyone putting effort in here that could be put into glimmer.

@tomdale c

packages/dom-helper/lib/main.js
krisselden: `process` is likely not something you want to expose on the sandbox context just for an is node check.

tomdale:
@stefanpenner @krisselden @chancancode Replaced detecting the environment with a feature detection. How does this approach look?

stefanpenner:
@tomdale I would feel better doing the test before the prototype member is set, rather then replacing the SharedFunction during invocation. Is this possible, or is it lazy fora specific reason?

packages/dom-helper/lib/main.js
```diff
@@ -565,15 +567,39 @@ prototype.parseHTML = function(html, contextualElement) {
...
+  var protocol = browserProtocolForURL.call(this, 'foobar:baz');
+  if (protocol === 'foobar:') {
+    // Swap in the method that doesn't do this test now that
+    // we know it works.
+    prototype.protocolForURL = browserProtocolForURL;
+  } else {
+    // Otherwise, we need to fall back to our own URL parsing.
+    // Global `require` is shadowed by Ember's loader so we have to use the fully
+    // qualified `module.require`.
+    URL = module.require('url');
```
stefanpenner: what if there is no require? Might be nice to fail fast then (might be aligned with my comment bellow about doing this before setting up the prototype.

packages/dom-helper/lib/main.js
```diff
@@ -565,15 +567,39 @@ prototype.parseHTML = function(html, contextualElement) {
...
+    prototype.protocolForURL = nodeProtocolForURL;
+  }
+
+  return this.protocolForURL(url);
+};
+
+function browserProtocolForURL(url) {
   if (!parsingNode) {
     parsingNode = this.document.createElement('a');
```
stefanpenner: we may as well just create this node in the outer context, rather then a branch hit once, its just extra weight on the inlining budget.

tomdale: I don't think you can create the node in the outer context because you don't yet have the `document`, which gets passed in by the consumer.

tomdale: Basically you aren't guaranteed to have the `document` until instantiation time.

stefanpenner: then it can be done in the constructor, and rather then overriding the method on the prototype, it could defer to a delegate which changes. This prevents some sadness we have been seeing with jits, and swapping out methods at runtime.


```js

function C() {
   this._prococalForURL = ...;
}

C.prototype.prococalForURL = function(url) {
  this._prococalForURL(url)
}
```

tomdale: What is the benefit of having a prototype method that delegates to an instance method that's been installed at construction time? Wouldn't it just be better to install the method on the instance?

```js
function C() {
  if (isNode) {
    this.protocolForURL = nodeProtocolForURL;
  } else {
    this.protocolForURL = browserProtocolForURL;
  }
}

function nodeProtocolForURL() { ... };
function browserProtocolForURL() { ... };
```

tomdale:
@chancancode @stefanpenner I updated the implementation to install the methods on the instance, and also added support for raw HTML sections. That last change let's us remove all of the monkeypatches from Ember/FastBoot.

tomdale:
Gonna merge this, @stefanpenner let me know if you've got any issues and I will fix.

PR #439, opened 2015-12-10
Use the global URL in node context if in sandbox

ronco:
This is for fastboot sandboxing.  URL is provided as a global to the
sandbox, and we'll use that rather than trying to require('url').

### DIFF
--- a/packages/dom-helper/lib/main.js
+++ b/packages/dom-helper/lib/main.js
@@ -1,4 +1,4 @@
-/*globals module*/
+/*globals module, URL*/
 
 import Morph from "./htmlbars-runtime/morph";
 import AttrMorph from "./morph-attr";
@@ -568,7 +568,7 @@ prototype.parseHTML = function(html, contextualElement) {
   return fragment;
 };
 
-var URL;
+var nodeURL = URL;
 var parsingNode;
 
 function installEnvironmentSpecificMethods(domHelper) {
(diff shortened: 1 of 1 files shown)


packages/dom-helper/lib/main.js
```diff
@@ -580,11 +579,16 @@ function installEnvironmentSpecificMethods(domHelper) {
...
     domHelper.protocolForURL = browserProtocolForURL;
+  } else if (typeof URL === 'object') {
+    // URL globally provided, likely from FastBoot's sandbox
+    domHelper.protocolForURL = nodeProtocolForURL;
   } else if (typeof module === 'object' && typeof module.require === 'function') {
     // Otherwise, we need to fall back to our own URL parsing.
     // Global `require` is shadowed by Ember's loader so we have to use the fully
     // qualified `module.require`.
+    /* jshint -W020 */
     URL = module.require('url');
```
ronco: I'm happy to rework this to use a differently named var in this scope to eliminate this warning.  That should work assuming the only place we need `URL` is in [nodeProtocolForURL](https://github.com/tildeio/htmlbars/pull/439/files#diff-9c4c7dd8b7ebbbd7f60a97eaaf81dde2L614).

tomdale: Hmm, we should probably not set the global `URL` here. Maybe we can alias it as a different variable that we assign the FastBoot-supplied `URL` to?

tomdale: Like `var nodeURL = URL; if (!nodeURL) { nodeURL = module.require('url'); }`

ronco: Yep, definitely along the lines of what I was thinking.  I'll rework this real quick.

ronco:
@tomdale reworked without assigning to the global `URL`.  I think there's still value in having the `module.require === "function"` protection before falling through to use `module.require()`.

tomdale:
@ronco Awesome, thank you! The SL test failed so I restarted. As soon as that passes I will merge this and cut a new release.

ronco:
Weird, looks like the IE9 SL test failed again.

PR #441, opened 2015-12-20
[BUGFIX beta] Range redirection

Serabe:
When an `range` calls `handleRedirect` it sets the `params` to an array
containing its value. This causes problems with contextual components called
with dot syntax.

If a contextual component is called with dot syntax and any kind of arguments
(either `params` and `attrs`) it gets compiled to `inline`. But when called
without any of them (`{{contextual.my-component}}`), it gets compiled to
a `range`.

The `range` calls to `handleRedirect` but sets `params` to be `[value]` instead
of an empty array, causing `positionalParams` to receive the wrong value.

Fixes emberjs/ember.js#12717

### DIFF
--- a/packages/htmlbars-runtime/lib/hooks.js
+++ b/packages/htmlbars-runtime/lib/hooks.js
@@ -835,7 +835,7 @@ export function partial(renderNode, env, scope, path) {
   that represents a range of content with a value.
 */
 export function range(morph, env, scope, path, value, visitor) {
-  if (handleRedirect(morph, env, scope, path, [value], {}, null, null, visitor)) {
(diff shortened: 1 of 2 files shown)


rwjblue:
Seems good to me. 

@mixonic - Any objections?

mixonic:
@Serabe hm, and you're tested this with Ember? I would expect the reason `value` is passed as `params` is to allow later keywords to add that stream as a dependency.

mixonic:
@rwjblue I think this will require a release of htmlbars and update to `beta` of Ember?

rwjblue:
Ya, thanks for the poke @mixonic. Released as v0.14.11.

PR #448, opened 2016-02-03
fixes emberjs/ember.js#12506

MatrixZ:


rwjblue:
This seems correct to me. 

@mixonic / @mmun - Any objections?

stefanpenner:
we should be careful about perf

chadhietala:
I second Stef's concern, `get` can be called a pretty significant amount of times.
screen shot 2016-02-02 at 11 45 29 pm

mmun:
We already lookup helpers for curlies that appear outside of attributes: https://github.com/tildeio/htmlbars/blob/master/packages/htmlbars-runtime/lib/hooks.js#L698. Doing it for attributes as well shouldn't affect perf that much? Besides, it is the correct thing to do.

That said, I don't like this particular implementation. `get` should not be looking up helpers. In `{{foo bar}}`, `bar` is never a helper. Instead, we should instead add an `ambiguous` hook that delegates `get` or `subexpr` depending on whether the path resolves to a helper or not. The `ambiguous` hook would be called whenever there was `{{foo}}` in the template. `{{foo.bar}}` should not trigger the `ambiguous` hook however.

PR #450, opened 2016-03-08
Fix loc info for content statements with leading newlines.

rwjblue:
Given the following:

```hbs
<div>
  {{#if isSingleError}}
    Single error here!
  {{else if errors}}
    Multiple errors here!
  {{else}}
    No errors found!
  {{/if}}
</div>
```

The `ElementNode` for the `<div>` was indicating that its `loc.end.column` was 12. This is ultimately due to how Handlebars handles trailing newlines after the `{{/if}}`. We were properly detecting that the node was `rightStripped` and adding the number of stripped newlines to the `tokenizer.line` number, but we did not reset the `tokenizer.column` also.

The fix is to ensure that the column is reset when incrementing the line count.

---

This is also being fixed in Glimmer in https://github.com/tildeio/glimmer/pull/91

### DIFF
--- a/packages/htmlbars-syntax/lib/parser/handlebars-node-visitors.js
+++ b/packages/htmlbars-syntax/lib/parser/handlebars-node-visitors.js
@@ -108,7 +108,9 @@ export default {
       changeLines = leadingNewlineDifference(content.original, content.value);
     }
 
-    this.tokenizer.line = this.tokenizer.line + changeLines;
+    this.tokenizer.line = content.loc.start.line + changeLines;
+    this.tokenizer.column = content.loc.start.column;
+
     this.tokenizer.tokenizePart(content.value);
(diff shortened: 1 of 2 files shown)


stefanpenner:
LGTM

rwjblue:
After digging a bit more, I realized that almost all `ElementNode` locations on the same line as a mustache were incorrect.  I have updated the approach to utilize the `ContentStatement`'s location information to reset the tokenizer's current positioning which fixes the following additional issues that I identified while digging into the `{{else if foo}}` issue:

* `{{#foo-bar}}<div>Foo</div>{{/foo-bar}}`  -- elements inside blocks had incorrect start/end
* `{{foo-bar}} <p>Hi!</p>` -- elements after a mustache had incorrect start/end

PR #455, opened 2016-04-21
Handle null/undefined url in nodeProtocolForURL

ronco:
The node version of protocolForURL was throwing an exception if the url is undefined [here](https://github.com/tildeio/htmlbars/blob/master/packages/dom-helper/lib/main.js#L619).  This is invoked in [sanitizeAttributeValue](https://github.com/tildeio/htmlbars/blob/master/packages/morph-attr/lib/sanitize-attribute-value.js#L47).  This was calling issues for bound attributes with no value in FastBoot environments.

This PR returns a sensible default (`:`) from `nodeProtocolForURL` when the `url` argument is a non-string.

I'm not sure how to test this though.  I'd like to get a test in place if possible before merging.

/cc @tomdale

### DIFF
--- a/packages/dom-helper/lib/main.js
+++ b/packages/dom-helper/lib/main.js
@@ -616,7 +616,10 @@ function browserProtocolForURL(url) {
 }
 
 function nodeProtocolForURL(url) {
-  var protocol = nodeURL.parse(url).protocol;
+  var protocol = null;
+  if (typeof url === 'string') {
+    protocol = nodeURL.parse(url).protocol;
+  }
   return (protocol === null) ? ':' : protocol;
 }
 


tomdale:
I think the best bet is to add a failing test to Ember.js. We can see that PR fail, then merge this fix and cut a new release, then see the PR pass and then merge the bumped version into Ember.

tomdale:
Fix looks good btw.

rwjblue:
Will also need to be fixed in tildeio/glimmer

rwjblue:
Still needs to be done in tildeio/glimmer, if we forget this it **will** regress...

ronco:
@rwjblue yes, @tomdale is going to follow up with @chancancode to get some direction there.

tomdale:
According to @chancancode the work has not yet started on the Node.js-compatible `DOMHelper`, what's the best way to track this?

PR #456, opened 2016-05-04
Fix comment and attribute start column.

rwjblue:
The EventedTokenizer only invokes the `beginComment` and `beginAttribute` methods **after** it has consumed the identifying chars.

Fixes https://github.com/rwjblue/ember-cli-template-lint/issues/89.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "htmlbars",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


rwjblue:
Also fixed in Glimmer here: https://github.com/tildeio/glimmer/pull/163

PR #463, opened 2016-05-13
Make setAttribute behave more like setProperty to align with glimmer

krisselden:


krisselden:
@chadhietala can you try this with your PR?

PR #464, opened 2016-05-13
Handle loc tracking locally.

rwjblue:
* Rely on simple-html-tokenizer calling `beginAttribute` correctly. Relies on changes in SHT to call `beginAttribute` with the proper `column` and `line` information.
* Use `tagOpen` hook to handle tag start loc tracking.
* Removes usage of `tagLine` and `tagColumn`.
* Track AttrNode value location information.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "htmlbars",
   "dependencies": {
-    "simple-html-tokenizer": "tildeio/simple-html-tokenizer#v0.2.4"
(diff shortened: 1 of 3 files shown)


rwjblue:
I still want to add loc info for attribute name and attribute value, but wanted to get the general 👍 on the changes to SHT first.

rwjblue:
Now tracking AttrNode value location info (with tests).

PR #466, opened 2016-05-24
Force `download` attribute on `<a>` tag to be treated as an attribute

pittst3r:
Addresses [emberjs/ember.js #13537](https://github.com/emberjs/ember.js/issues/13537).

[`normalizeProperty`](https://github.com/tildeio/htmlbars/blob/master/packages/dom-helper/lib/prop.js#L11) interprets the `download` attribute as a property but, in my estimation, we want to treat it as an attribute, so I've added it to the [overrides object](https://github.com/tildeio/htmlbars/blob/master/packages/dom-helper/lib/prop.js#L40). If I am incorrect in this I can close this and update the original issue referenced above.

### DIFF
--- a/packages/dom-helper/tests/prop-test.js
+++ b/packages/dom-helper/tests/prop-test.js
@@ -3,9 +3,10 @@ import { normalizeProperty } from 'dom-helper/prop';
 QUnit.module('dom-helper prop');
 
 test('type.attr, for element props that for one reason or another need to be treated as attrs', function() {
-  expect(13);
+  expect(14);
 
   [
+    { tagName: 'A',        key: 'download' },
     { tagName: 'TEXTAREA', key: 'form' },
     { tagName: 'BUTTON',   key: 'type' },
     { tagName: 'INPUT',    key: 'type' },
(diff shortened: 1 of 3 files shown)


packages/dom-helper/tests/prop-test.js
```diff
@@ -3,17 +3,18 @@ import { normalizeProperty } from 'dom-helper/prop';
...
+    { tagName: 'A',        key: 'download' },
     { tagName: 'TEXTAREA', key: 'form' },
     { tagName: 'BUTTON',   key: 'type' },
     { tagName: 'INPUT',    key: 'type' },
     { tagName: 'INPUT',    key: 'list' },
     { tagName: 'INPUT',    key: 'form' },
     { tagName: 'INPUT',    key: 'autocorrect' },
     { tagName: 'OPTION',   key: 'form' },
-    { tagName: 'INPUT',    key: 'form' },
+    { tagName: 'SELECT',   key: 'form' },
```
robbiepitts: `{ tagName: 'INPUT',    key: 'form' }` was duplicated here and `{ tagName: 'SELECT',   key: 'form' }` was not present but should have been. Unrelated to the issue but I figured I'd fix it while I was in here.

robbiepitts:
Closing this since glimmer 2 has been released and there's been no action on this PR.

PR #469, opened 2016-06-17
Ensure statement nodes are the same length.

### LINKED ISSUE #468: Make all nodes in expressions a fixed length
The `acceptExpression` method of the htmlbars runtime is a source of a lot of V8 JIT churn because of BoundsCheck hoisting, if we made all the nodes the same length, just filling the empty slots with undefined, when the acceptParams and acceptHash loops caused the argument destructing BoundsCheck opcode to be hoisted outside the loop it would not cause a deopt because all nodes would be the same length.

rwjblue:
Fixes #468.

### DIFF
--- a/packages/htmlbars-runtime/lib/expression-visitor.js
+++ b/packages/htmlbars-runtime/lib/expression-visitor.js
@@ -36,10 +36,12 @@ function acceptExpression(node, env, scope) {
 
   // Primitive literals are unambiguously non-array representations of
   // themselves.
-  if (typeof node !== 'object' || node === null) {
-    ret.value = node;
-  } else {
+  if (Array.isArray(node)) {
+    // if (node.length !== 7) { throw new Error('FIXME: Invalid statement length!'); }
+
     ret.value = evaluateNode(node, env, scope);
+  } else {
+    ret.value = node;
   }
 
   return ret;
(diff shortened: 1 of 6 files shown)


packages/htmlbars-runtime/lib/expression-visitor.js
```diff
@@ -36,10 +36,12 @@ function acceptExpression(node, env, scope) {
 
   // Primitive literals are unambiguously non-array representations of
   // themselves.
-  if (typeof node !== 'object' || node === null) {
-    ret.value = node;
-  } else {
+  if (Array.isArray(node)) {
+    // if (node.length !== 7) { throw new Error('FIXME: Invalid statement length!'); }
```
rwjblue: This was used to ensure that both HTMLBars and Ember's test suites pass without this error being thrown.  Ideally, we would include this as an assertion in the build output, but there is no machinery setup to allow that to be stripped at the moment.

mixonic:
review plz @krisselden?

krisselden:
@mixonic it looks good, @rwjblue had said he wanted to test it, I gave him some instructions to do this.

This does look like it would address the issue in #468

rwjblue:
## Before:

before trace-deopt.sh output

ir hydra output

## After:

after trace-deopt.hs output

ir hyrdra output

rwjblue:
I was being trolled by caching in the *After:* screenshots above.

@krisselden worked with me to confirm this is now properly avoiding a deopt.

screenshot

